PR #485, opened 2025-12-23
Fix send flow UI issues on Android and add validation toasts

praveenperera:
## Summary
- Fix dark mode contrast for Swipe to Send by adding `swipeTrackBg` color matching iOS `systemGray5` (#E5E5EA light, #2C2C2E dark)
- Prevent amount text overlap with sats dropdown by adding horizontal padding to match iOS
- Show snackbar (Android) / FloaterPopup (iOS) on validation failure when field has content, dismissing after 1 second

## Test plan
- [ ] Test Swipe to Send in dark mode on Android - circle and track should have clear contrast
- [ ] Enter many digits in amount field - text should not overlap with "sats" dropdown
- [ ] Tap Next with invalid address (has content) - should show validation toast
- [ ] Tap Next with empty address - should focus address field without toast
- [ ] Verify iOS floater popup appears and dismisses after 1 second

## Summary by CodeRabbit

* **Style**
  * Added horizontal padding to the amount input for improved spacing.
  * Updated swipe-to-send track background to better match light/dark themes and added corresponding theme colors.

* **Bug Fixes**
  * Validation now presents contextual inline error messages only when a field contains invalid content.
  * Fields that fail validation are automatically focused to guide correction.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/ios/Cove/Flows/SendFlow/SendFlowSetAmountScreen.swift
+++ b/ios/Cove/Flows/SendFlow/SendFlowSetAmountScreen.swift
@@ -65,27 +65,29 @@ struct SendFlowSetAmountScreen: View {
     // MARK: Actions
 
     private func next() {
-        guard performValidation() else { return }
-        sendFlowManager.dispatch(action: .finalizeAndGoToNextScreen)
+        Task {
+            guard await performValidation() else { return }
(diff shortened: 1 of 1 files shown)


praveenperera:
@greptile-apps re-review

PR #553, opened 2026-02-03
Show complementary amount instead of block number in transaction rows

praveenperera:
Display fiat amount when in BTC mode, and BTC/sats amount when in fiat mode as the secondary line in transaction rows on both platforms

## Summary by CodeRabbit

* **New Features**
  * Transaction cards show a secondary amount (BTC or fiat) beneath the primary amount for confirmed, unconfirmed, and unsigned transactions; confirmed items use this secondary currency in place of block height. Privacy masking preserved.
  * Amounts now display direction-aware formatting (sent/received) for both BTC and fiat secondary values.

* **Bug Fixes**
  * Deleting an unsigned transaction surfaces a user-facing alert when it fails.


praveenperera:
@greptile-apps re-review

PR #625, opened 2026-04-14
fix(tap-card): add type-safe ConnectionLost error for NFC tag drops

### LINKED ISSUE #439: Replace string-based connection error checking with type-safe error handling
## Problem

Currently, both iOS and Android use fragile string-based checking for connection errors:

**Android** (`TapCardNfcManager.kt:160-161`):
```kotlin
e.message?.contains("connection lost") == true ->
    "Tag connection lost, please hold your phone still"
```

**iOS** (`TapSignerNFC.swift:373-376`):
```swift
case .readerTransceiveErrorTagConnectionLost:
    tapSignerError = .Unknown("Tag connection lost, please hold your phone still")
```

This pattern is fragile because:
- Depends on message formatting staying consistent
- No compile-time safety
- Can break with error format changes

## Solution

Implement type-safe error checking similar to the existing `tapSignerErrorIsAuthError()` pattern:

### Changes Required

1. **Update Rust `TransportError` enum** (`rust/src/tap_card.rs`)
   - Add `ConnectionError(String)` variant to distinguish connection/IO errors

2. **Modify iOS transport layer** (`ios/Cove/TapSignerNFC.swift`)
   - Throw `TransportError.ConnectionError` instead of `UnknownError` for tag connection issues

3. **Modify Android transport layer** (`android/app/src/main/java/org/bitcoinppl/cove/nfc/TapCardNfcManager.kt`)
   - Throw `TransportError.ConnectionError` instead of generic Exception

4. **Add helper method in Rust** (`rust/src/tap_card/tap_signer_reader.rs`)
   ```rust
   pub fn is_connection_error(&self) -> bool {
       matches!(
           self,
           TapSignerReaderError::TapSignerError(TransportError::ConnectionError(_))
       )
   }
   ```

5. **Export via UniFFI**
   - Add to uniffi exports like existing `tap_signer_error_is_auth_error`

6. **Update iOS** to use `tapSignerErrorIsConnectionError()`

7. **Update Android** to use `tapSignerErrorIsConnectionError()`

8. **Rebuild rust bindings** for both platforms

## Benefits

- Type-safe error handling across both platforms
- Consistent with existing `tapSignerErrorIsAuthError()` pattern
- Won't break if error message formatting changes
- Compile-time safety

## Related

This follows the same pattern as PR #433 which replaced string-based BadAuth checking with type-safe error handling.

Thunder-Blaze:
### Description

NFC tag disconnections during APDU communication are reported as `TransportError::UnknownError` / `TransportException.UnknownException`. Callers have no type-safe way to distinguish a connection drop from other unknown errors, making retry logic fragile (string matching on error messages).

### Issue
- Fixes #439 

### Changes

- Add `ConnectionLost(String)` variant to `TransportError` (Rust)
- Map `ConnectionLost` through `From<TransportError> for ApduError`
- Add `is_connection_error()` to `TapSignerReaderError` + UniFFI export `isConnectionError()`
- **Android**: `TapCardTransport.transmitApdu` throws `ConnectionLostException` instead of `UnknownException`
- **iOS**: `readerTransceiveErrorTagConnectionLost` maps to `.ConnectionLost` instead of `.Unknown`

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/nfc/TapCardNfcManager.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/nfc/TapCardNfcManager.kt
@@ -260,7 +260,7 @@ private class TapCardTransport(
             response
         } catch (e: Exception) {
             Log.e(tag, "APDU error", e)
-            throw TransportException.UnknownException("Tag connection lost, please hold your phone still")
+            throw TransportException.ConnectionLostException("Tag connection lost, please hold your phone still")
         }
     }
 }
--- a/ios/Cove/TapSignerNFC.swift
+++ b/ios/Cove/TapSignerNFC.swift
@@ -394,7 +394,7 @@ private class TapCardNFC: NSObject, NFCTagReaderSessionDelegate {
         case let .some(error):
             switch error.code {
             case .readerTransceiveErrorTagConnectionLost:
-                tapSignerError = .Unknown("Tag connection lost, please hold your phone still")
+                tapSignerError = .ConnectionLost("Tag connection lost, please hold your phone still")
                 session.invalidate(
                     errorMessage: "Tag connection lost, please hold your phone still"
                 )
--- a/rust/src/tap_card.rs
+++ b/rust/src/tap_card.rs
@@ -25,6 +25,9 @@ pub enum TransportError {
     #[error("CvcChangeError: {0}")]
     CvcChangeError(String),
 
+    #[error("ConnectionLost: {0}")]
+    ConnectionLost(String),
+
     #[error("UnknownError: {0}")]
     UnknownError(String),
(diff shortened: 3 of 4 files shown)


praveenperera:
make it work first @Thunder-Blaze just `just ci` locally before asking for review

Thunder-Blaze:
yes, I'm really sorry about that
I should've made a Draft PR instead of a PR

I was in the middle of it when I noticed that changing the Error enum in rust-cktap would improve the overall typing
So I made a PR for it today but Coderabbit commented that it is gonna affect downstream crates which is also very true, so I'm kinda confused

Which way do you think I should implement @praveenperera ?

praveenperera:
it should go in the upstream rust-cktap

PR #630, opened 2026-04-14
Extract pj and pjos params from BIP 21 URIs

harsh04044:
Switched to PDK's payjoin::Uri for BIP21 parsing instead of manually extracting query params. This is the foundation for sender-side PayJoin support, no negotiation or UI yet.

pj and pjos are extracted through PDK's parser and carried through AddressWithNetwork to the FFI layer. Invalid pj URLs now reject the whole URI per BIP78. pjos defaults to enabled when omitted, matching PDK semantics. Removed the url crate from cove-types since it's no longer needed.

## Summary by CodeRabbit

* **New Features**
  * Support for Payjoin metadata in Bitcoin payment URIs (pj and pjos), exposed via new public accessors (including Kotlin bindings).

* **Refactor**
  * URI parsing now returns a richer parsed result so amount and Payjoin fields reliably propagate into address objects.

* **Tests**
  * Added tests for pj/pjos parsing, malformed pj, invalid pjos values, and pjos-without-pj rejection.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -57,6 +57,7 @@ bitcoin = { version = "0.32" }
 bdk_wallet = { version = "2.3", features = ["keys-bip39", "file_store", "rusqlite"] }
 bip39 = { version = "2.2", features = ["zeroize"] }
 bip329 = { version = "0.4" }
+payjoin = { version = "0.25", default-features = false, features = ["v1"] }
 pubport = { version = "0.5" }
 
 # bitcoin nodes
--- a/rust/crates/cove-types/src/address.rs
+++ b/rust/crates/cove-types/src/address.rs
@@ -14,9 +14,9 @@ use bitcoin::{
     address::{NetworkChecked, NetworkUnchecked},
     params::Params,
 };
+use payjoin::UriExt as _;
 use serde::Deserialize;
 use strum::IntoEnumIterator as _;
-use url::Url;
 
 use crate::{Network, amount::Amount, transaction::TransactionDirection};
 
@@ -211,53 +211,35 @@ fn parse_bitcoin_uri(input: &str) -> Result<ParsedBitcoinUri, Error> {
         return Err(Error::EmptyAddress);
     }
 
-    let has_scheme =
-        input.split_once(':').is_some_and(|(scheme, _)| scheme.eq_ignore_ascii_case("bitcoin"));
-
-    let normalized = if has_scheme { input.to_string() } else { format!("bitcoin:{input}") };
-    let url = Url::parse(&normalized).map_err(|_| Error::InvalidAddress)?;
-
-    if !url.scheme().eq_ignore_ascii_case("bitcoin") {
-        return Err(Error::InvalidAddress);
-    }
-
-    let address = match (url.path(), url.host_str()) {
-        (address, None) | ("", Some(address)) if !address.is_empty() => String::from(address),
-        _ => {
-            return Err(Error::EmptyAddress);
+    // Normalize: strip "bitcoin://" to "bitcoin:" and prepend scheme if missing
+    let normalized = match input.split_once(':') {
+        Some((scheme, rest)) if scheme.eq_ignore_ascii_case("bitcoin") => {
+            let rest = rest.strip_prefix("//").unwrap_or(rest);
+            format!("bitcoin:{rest}")
         }
+        _ => format!("bitcoin:{input}"),
     };
 
-    // take amount from query params
-    let amount = url.query_pairs().find(|(key, _)| key == "amount").and_then(|(_, value)| {
-        let value = value.trim();
-        value.parse::<f64>().ok().and_then(|btc| Amount::from_btc(btc).ok())
-    });
+    let uri = payjoin::Uri::try_from(normalized.as_str()).map_err(|_| Error::InvalidAddress)?;
 
-    // payjoin endpoint
-    let pj = url
-        .query_pairs()
-        .find(|(key, _)| key == "pj")
-        .map(|(_, value)| value.to_string())
-        .filter(|v| !v.is_empty())
-        .and_then(|v| Url::parse(&v).ok().map(|_| v));
+    let amount = uri.amount.map(|a| Amount::from_sat(a.to_sat()));
 
-    // payjoin output substitution — per PDK/rust-payjoin:
-    // pjos=0 -> Some(false), pjos=1 -> Some(true), other values -> error, omitted -> None
-    let pjos = url
-        .query_pairs()
-        .find(|(key, _)| key == "pjos")
-        .map(|(_, value)| match value.trim() {
-            "0" => Ok(false),
-            "1" => Ok(true),
-            _ => Err(Error::InvalidAddress),
-        })
-        .transpose()?;
+    // skip network check; try_new() validates later
+    let checked = uri.assume_checked();
+    let address = checked.address.to_string();
 
-    // pjos without pj is a malformed URI
-    if pjos.is_some() && pj.is_none() {
-        return Err(Error::InvalidAddress);
-    }
+    // Extract payjoin params if present
+    let (pj, pjos) = match checked.check_pj_supported() {
+        Ok(pj_uri) => {
+            let endpoint = pj_uri.extras.endpoint();
+            let pjos = match pj_uri.extras.output_substitution() {
+                payjoin::OutputSubstitution::Enabled => Some(true),
+                payjoin::OutputSubstitution::Disabled => Some(false),
+            };
+            (Some(endpoint), pjos)
+        }
+        Err(_) => (None, None),
+    };
 
     Ok(ParsedBitcoinUri { address, amount, pj, pjos })
 }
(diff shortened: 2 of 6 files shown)


rust/crates/cove-types/src/address.rs
```diff
@@ -227,7 +236,19 @@ fn parse_bitcoin_uri(input: &str) -> Result<(String, Option<Amount>), Error> {
...
     });
 
-    Ok((address, amount))
+    // payjoin endpoint
+    let pj = url
+        .query_pairs()
+        .find(|(key, _)| key == "pj")
+        .map(|(_, value)| value.to_string())
+        .filter(|v| !v.is_empty())
+        .and_then(|v| Url::parse(&v).ok().map(|_| v));
```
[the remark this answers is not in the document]
harsh04044: Intentionally keeping HTTPS enforcement out of the parser. BIP 78 allows .onion HTTP endpoints for Tor, and test environments use http://localhost. HTTPS enforcement belongs in the PayJoin sender layer when the actual request is made, not at URI parse time.

rust/crates/cove-types/src/address.rs
```diff
@@ -60,6 +60,16 @@ pub struct AddressWithNetwork {
...
+    pub pj: Option<String>,
+    pub pjos: Option<bool>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct ParsedBitcoinUri {
+    address: String,
+    amount: Option<Amount>,
+    pj: Option<String>,
+    pjos: Option<bool>,
```
praveenperera: good, thanks for extracting this out

rust/crates/cove-types/src/address.rs
```diff
@@ -167,24 +177,23 @@ impl AddressWithNetwork {
...
-            return Ok(Self { address: address.into(), network, amount });
-        }
-
-        let network = Network::Testnet;
-        if let Ok(address) = address.clone().require_network(network.into()) {
-            return Ok(Self { address: address.into(), network, amount });
-        }
+            parsed.address.parse().map_err(|_| Error::InvalidAddress)?;
+
+        let build = |address: Address, network: Network| Self {
```
praveenperera: why a closure here?

rust/crates/cove-types/src/address.rs
```diff
@@ -167,24 +177,23 @@ impl AddressWithNetwork {
...
+            network,
+            amount: parsed.amount,
+            pj: parsed.pj.clone(),
+            pjos: parsed.pjos,
+        };
 
-        let network = Network::Signet;
-        if let Ok(address) = address.require_network(network.into()) {
-            return Ok(Self { address: address.into(), network, amount });
+        for network in [Network::Bitcoin, Network::Testnet, Network::Signet] {
```
praveenperera: use Network::iter

rust/crates/cove-types/src/address.rs
harsh04044: Good point, dropped it.

rust/crates/cove-types/src/address.rs
harsh04044: Done

praveenperera:
@greptile-apps re-review

harsh04044:
Greptile seems stuck.

praveenperera:
@harsh04044 did you run just bi && just ba to rebuild bindings?

harsh04044:
> @harsh04044 did you run just bi && just ba to rebuild bindings?

Ran just bi and just ba, both pass. Pushed the regenerated bindings.
image
image

rust/crates/cove-types/src/address.rs
```diff
@@ -227,7 +234,22 @@ fn parse_bitcoin_uri(input: &str) -> Result<(String, Option<Amount>), Error> {
...
     });
 
-    Ok((address, amount))
+    // payjoin endpoint
+    let pj = url
+        .query_pairs()
+        .find(|(key, _)| key == "pj")
+        .map(|(_, value)| value.to_string())
+        .filter(|v| !v.is_empty())
+        .and_then(|v| Url::parse(&v).ok().map(|_| v));
```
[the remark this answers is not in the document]
harsh04044: @coderabbitai Keeping scheme enforcement out of the parser intentionally. BIP 78 allows .onion endpoints over HTTP for Tor, and test setups commonly use http://localhost. Scheme validation belongs in the PayJoin sender layer when the request is actually made, not at URI parse time.

harsh04044:
CI clippy is failing on master-side lints in `historical_price_service.rs`:74 and `coin_control_manager/state.rs`:104,130. Not from my changes. Happy to fix in this PR or a separate one, whichever you prefer.
cc: @praveenperera

praveenperera:
@harsh04044 

This pjos parsing is too narrow if the goal is broad Payjoin compatibility. Right now it assumes BIP78-only semantics by treating only pjos=0 as meaningful, accepting pjos without pj, and ignoring pjos=1/other values. That does not match current Payjoin Dev Kit v2 behavior.

  In PDK / rust-payjoin:

  - pjos=0 means output substitution disabled
  - pjos=1 means output substitution enabled
  - omitted pjos defaults to enabled
  - pjos without pj is rejected as MissingEndpoint
  - other pjos values are parse errors

  References:

  - PDK URI parsing: https://github.com/payjoin/rust-payjoin/blob/master/payjoin/src/core/uri/mod.rs
  - v2 pj parameter structure: https://github.com/payjoin/rust-payjoin/blob/master/payjoin/src/core/uri/v2.rs
  - v2 sender behavior/tests: https://github.com/payjoin/rust-payjoin/blob/master/payjoin/src/core/send/v2/mod.rs
  - BIP 77 async Payjoin: https://bips.dev/77/

  Because this branch is adding generic payjoin URI extraction, I think we should align with PDK-compatible semantics here rather than bake in a BIP78-only interpretation. The tests below also lock in the incompatible behavior:

  - test_parse_bitcoin_uri_with_pjos_only
  - test_parse_bitcoin_uri_pjos_non_zero_is_none

rust/crates/cove-types/src/address.rs
```diff
@@ -227,7 +234,22 @@ fn parse_bitcoin_uri(input: &str) -> Result<(String, Option<Amount>), Error> {
...
+
+    // payjoin output substitution — per BIP 78, only pjos=0 is defined (disables substitution)
+    let pjos = url
+        .query_pairs()
+        .find(|(key, _)| key == "pjos")
+        .filter(|(_, value)| value.trim() == "0")
+        .map(|_| false);
```
Sandipmandal25: the pjos logic only handles pjos=0 and silently drops everything else but pjos=1 is also a valid defined value in PDK v2, it means output substitution is enabled so it should map to Some(true) not None. also other invalid values should be an error not silent None. and if pjos is present but pj is missing the URI is
malformed, PDK rejects it as MissingEndpoint, we should too

rust/crates/cove-types/src/address.rs
```diff
@@ -590,4 +620,66 @@ mod tests {
...
+    fn test_parse_bitcoin_uri_with_pj_and_pjos_disabled() {
+        let a = "bitcoin:bc1q00000002ltfnxz6lt9g655akfz0lm6k9wva2rm?amount=0.001&pj=https://example.com/payjoin&pjos=0";
+        let parsed = parse_bitcoin_uri(a).unwrap();
+        assert_eq!(parsed.pj.as_deref(), Some("https://example.com/payjoin"));
+        assert_eq!(parsed.pjos, Some(false));
+    }
+
+    #[test]
+    fn test_parse_bitcoin_uri_with_pjos_only() {
+        let a = "bitcoin:bc1q00000002ltfnxz6lt9g655akfz0lm6k9wva2rm?pjos=0";
```
Sandipmandal25, line 670: this test accepts pjos=0 without a pj endpoint which is wrong. pjos without pj is a malformed URI per PDK. the test is locking in wrong behavior

rust/crates/cove-types/src/address.rs
```diff
@@ -590,4 +620,66 @@ mod tests {
...
+        let a = "bitcoin:bc1q00000002ltfnxz6lt9g655akfz0lm6k9wva2rm?pj=not-a-url";
+        let parsed = parse_bitcoin_uri(a).unwrap();
+        assert_eq!(parsed.pj, None);
+    }
+
+    #[test]
+    fn test_parse_bitcoin_uri_pjos_non_zero_is_none() {
+        // per BIP 78, only pjos=0 is defined; other values should be treated as absent
+        let a = "bitcoin:bc1q00000002ltfnxz6lt9g655akfz0lm6k9wva2rm?pjos=1";
+        let parsed = parse_bitcoin_uri(a).unwrap();
```
Sandipmandal25: pjos=1 is not an unknown value, it means output substitution is enabled so the assertion should be Some(true) not None. if you want to test invalid values use pjos=garbage or something that is actually undefined

Sandipmandal25 (changes requested):
the pjos=1 case seems wrong based on how PDK v2 handles it, left some inline comments. @praveenperera  what do you think does this match what you had in mind?

praveenperera:
ya i think your comments make sense, fix those and this can merge, thank you

harsh04044:
> ya i think your comments make sense, fix those and this can merge, thank you

Updated to match PDK semantics. pjos=0 maps to disabled, pjos=1 to enabled, other values are parse errors, pjos without pj is rejected. Rebuilt bindings. 
Thanks for the references.

praveenperera:
@harsh04044 we decided to change this to use the parser from pdk right?

harsh04044:
> @harsh04044 we decided to change this to use the parser from pdk right?

Yeah, updating this to use PDK's parser. Will push shortly.

praveenperera:
thanks no rush, just trying to get all the PRs reviewed

harsh04044:
Switched to PDK's URI parser, payjoin::Uri handles all BIP21 parsing now, removed the hand-rolled pj/pjos extraction and the url crate dependency.

harsh04044:
@praveenperera also shared a competency test draft on Discord, would appreciate your feedback when you get a chance

rust/crates/cove-types/src/address.rs
praveenperera: maybe we could combine this into an enum? or a struct, preferably and enum, would that make sense?

praveenperera (changes requested):

rust/crates/cove-types/src/address.rs
harsh04044: Yeah that makes sense.

harsh04044:
Two options I see:

```rust
pub struct PayJoinParams {
    pub endpoint: String,
    pub output_substitution: bool,
}
```

Use `Option<PayJoinParams>` in `AddressWithNetwork`. That way `pjos` can't exist without `pj`.

Or if we want to distinguish versions later:

```rust
pub enum PayJoinVersion {
    V1 { endpoint: String, output_substitution: bool },
    V2 { endpoint: String, output_substitution: bool },
}
```

Wrapped in `Option<PayJoinVersion>`. Could be useful when we need to dispatch to different sender flows.

praveenperera:
@harsh04044 since we are only doing payjoin v2, i dont think we need the version with `V1`

I think i'm just wondering if we should do 

```rust
pub struct PayJoinParams {
    pub endpoint: String,
    pub output_substitution: bool,
}
```

or somethign like this 

```rust
pub enum PayjoinParams {
  WithOutputSubstitution(String)
  NoSubstitution(String)
}
```

harsh04044:
Good call, I'll go with the struct:

```rust
pub struct PayJoinParams {
    pub endpoint: String,
    pub output_substitution: bool,
}
```

That way `pjos` can't exist without `pj` and we keep the output substitution info from the URI. I'll update the PR.

praveenperera:
@harsh04044 why do you think the struct is better than the enum?

harsh04044:
Honestly, mainly simplicity. The struct is one type with two fields, easy to extend if we need to add more params later (like relay hints or expiration). With the enum, adding a new field means adding it to both variants.

But your enum is arguably more readable , `WithOutputSubstitution` vs `NoSubstitution` makes the meaning obvious without checking a bool value.

I don't have a strong preference either way. Happy to go with whichever you think fits Cove's patterns better.

praveenperera:
I know this seems like a small thing, and I don’t want to spend too much time on it, but getting the data structures right early on can make things much easier later.

For your first concern, we can use a shared struct to hold the data. I guess this depends on how much output substitution affects how it’s used. The advantage of using an enum here is that each branch of the enum could have a different struct, which gives us type safety at compile time.

So this could make sense if, with substitution or without substitution, is used in completely different ways and we have to make sure, for safety reasons, not to interchange them. I’m not familiar with Payjoin enough to know if this makes sense or not. I think we need to do a little bit more research.

harsh04044:
@praveenperera , looked into how `output_substitution` flows through rust-payjoin. Cove passes the full `PjUri` to `SenderBuilder::new(psbt, pj_uri)`. PDK handles output substitution internally in two places:

1. Request body serialization (tells the receiver the preference)
2. `process_proposal()` validation (rejects proposals that violate it)

Cove never branches on it. We get back either a validated PSBT or an error regardless of which setting was used. So the enum would force pattern matching at every endpoint access, but both branches would do the same thing in our code.

Given that, I think the struct makes more sense, happy to go with the enum if you see a reason I'm missing tho.

praveenperera:
makes sense lets do the struct, thanks for doing the research @harsh04044

harsh04044:
> makes sense lets do the struct, thanks for doing the research @harsh04044

Done, pushed the PayJoinParams struct refactor, rebuilt binidings

rust/crates/cove-types/src/address.rs
```diff
@@ -198,36 +208,43 @@ impl AddressWithNetwork {
...
+    // Normalize: strip "bitcoin://" to "bitcoin:" and prepend scheme if missing
+    let normalized = match input.split_once(':') {
+        Some((scheme, rest)) if scheme.eq_ignore_ascii_case("bitcoin") => {
+            let rest = rest.strip_prefix("//").unwrap_or(rest);
+            format!("bitcoin:{rest}")
+        }
+        _ => format!("bitcoin:{input}"),
+    };
+
+    let uri = payjoin::Uri::try_from(normalized.as_str()).map_err(|_| Error::InvalidAddress)?;
```
praveenperera: don't fail entire address parsing on payjoin fail

praveenperera (changes requested):
window.rust/crates/cove-types/src/address.rs:226-245 now makes AddressWithNetwork::try_new() fail whenever payjoin::Uri::try_from(...) rejects a pj parameter, but rust/src/manager/send_flow_manager.rs:1490-1523 still ignores address_with_network.payjoin entirely and only uses the parsed address plus amount. So Cove is still effectively a non-payjoin sender, yet malformed/insecure pj values now turn an otherwise payable invoice into InvalidAddress. BIP78’s backward-compatibility section says non-payjoin senders should ignore pj and proceed with a normal payment: https://bips.xyz/78#backward-compatibility . Until the send path actually consumes payjoin, this parser should treat pj as best-effort instead of fatal.

So even later if payjoin parsing fails, just parse as a non payjoin Addreess

rust/crates/cove-types/src/address.rs
harsh04044: Fixed. Malformed pj= values now fall back to parsing as a regular address with payjoin: None, per BIP 78 backward compatibility. Added a test for the fallback path.

rust/crates/cove-types/src/address.rs
```diff
@@ -55,11 +55,25 @@ pub struct AddressInfoWithDerivation {
     pub derivation_path: Option<String>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, uniffi::Record)]
+pub struct PayJoinParams {
+    pub endpoint: String,
+    pub output_substitution: bool,
```
praveenperera: one more small thing, should we call this `output_substituation_enabled` to make sure theres 0 confusion?

praveenperera (changes requested):
one more small change, will merge after

rust/crates/cove-types/src/address.rs
harsh04044: Makes sense, renaming to output_substitution_enabled.

harsh04044: Done, renamed to output_substitution_enabled, rebuilt binidings.

praveenperera (approved):
looks good thanks @harsh04044 approved!

rust/Cargo.toml
```diff
@@ -57,6 +57,7 @@ bitcoin = { version = "0.32" }
 bdk_wallet = { version = "2.3", features = ["keys-bip39", "file_store", "rusqlite"] }
 bip39 = { version = "2.2", features = ["zeroize"] }
 bip329 = { version = "0.4" }
+payjoin = { version = "0.25", default-features = false, features = ["v1"] }
```
DanGould, line 60: Why use v1?

Sandipmandal25: v1 just to prove the signing pipeline works same psbt mechanics as v2. actual impl is targeting bip77/v2.

DanGould: Interesting. I think you'll find the v2 sender just as simple as the v1 sender if you write a no-op persister or use the provided in-memory persister (ignoring persistence) for your testing, and have a prototype that's a closer shape to production

Sandipmandal25: yeah already moving to v2 with proper session persistence over coves db. good tip on the in memory persister for testing before the real one is wired up.

Issue #146, opened 2024-12-17
Wipe wallet data after 10 failed attempts


praveenperera:

Justxd22:
Hello, are you planning to support this? If yes please assign! 

im from SoB :]

PR #631, opened 2026-04-15
Fix PSBT export share sheet on hardware send flow

### LINKED ISSUE #313: Export Transaction - Save to File isn’t working
On iPad, after having created a transaction, QR Code and NFC work but when I click on more, and then Save to files , it fails to open my files and allow me to save the psbt.

### LINKED ISSUE #449: Export Transaction fails (Testnet4)
Attempting to export a transaction to a PSBT file from a watch-only wallet for the purpose of signing externally fails to generate the PSBT file.

To demonstrate, I created 2 new Testnet4 wallets from within Sparrow wallet; and funded one of them.  Then, I added both wallets to Cove as watch-only wallets.
From the funded wallet I created a transaction to send funds to the unfunded wallet.  Then, I attempted Export Transaction to create a PSBT so that I could sign it externally (was going to be Sparrow); but Cove failed to create the file.

A more detailed process of what I did follows below.  I included the tpubs of the wallets I used so that you can try to recreate the error with the same wallets I used.   Let me know if I haven't described something in enough detail. I have many more screenshots of me stepping through the process that I can add if necessary.  And, let me know if there's any more info I can provide that you think might help.

Thank you!

PS. I  get the same results if I import the watch-only wallets by scanning the QRs for them from Sparrow instead of copy/pasting the tpubs.

1. Settings -> Network -> Testnet4
2. Copy sending wallet tpub: tpubDDGvQAWv4GhRAjisKnJpsBTteCrRsz9KiwtGQrvys6NtAEeyrHNPNfcfazdbNSXMHomZeukPL5DjB8d2sSC8nEocjnV9JyNZBxuTkgQiMkd
3. Add Wallet -> Hardware Wallet -> Paste
4. Wallet Settings -> Name -> 'Cove-send'
5. Copy receiving wallet tpub: tpubDCt9bbHtF7QUHD3hUdRvubL6AREG31QUC2gHCXkqLifvuJenXHZa9eZFhJbDyNCNAWwhXL5hxFvYjNuS54baT8kVXJQT6dAdnqZc1uNEdYW
6. Add Wallet -> Hardware Wallet -> Paste
7. Wallet Settings -> Name -> 'Cove-receive'
8. from 'Cove-recieve' wallet Receive -> Copy Address
9. from 'Cove-send' wallet Send -> 10,000 sats -> Paste receive address copied from step 8 -> Next

Image

10. Export Transaction -> More...
11. ...

Choosing Signal results in error message:

Image

 
Choosing Save to Files results in the screen blanking momentarily before returning to the same options (with no additional message):

Image

pradhyum6144:
Replace the SwiftUI ShareLink inside the export-transaction confirmationDialog with an imperative UIActivityViewController call, deferred by 400ms so the dialog's dismissal animation can finish before the share sheet presents.

Presenting a ShareLink directly from a confirmationDialog action races the UIAlertController dismissal: Save to Files fails silently and share extensions like Signal surface a generic error. The deferred call goes through the existing ShareSheet helper (already used by MoreInfoPopover.exportTransactions) which wraps UIActivityViewController with proper iPad popover anchoring and temp-file cleanup.

Adds a Data variant of ShareSheet.present for binary payloads and drops the now-unused PSBTFile Transferable struct. iOS only; Android's FileProvider + ACTION_SEND path is unaffected.

Closes #449
Closes #313

## Release Notes

* **Bug Fixes**
  * Improved error handling for PSBT file exports with explicit failure notifications.

* **Refactor**
  * Enhanced file sharing mechanism with more robust and reliable write operations.

### DIFF
--- a/ios/Cove/PSBTFile.swift
+++ b/ios/Cove/PSBTFile.swift
@@ -5,30 +5,7 @@
 //  Created by Praveen Perera on 11/24/24.
 //
 import Foundation
-import SwiftUI
 import UniformTypeIdentifiers
-
-struct PSBTFile: Transferable {
-    let data: Data
-    let filename: String
-
-    static var transferRepresentation: some TransferRepresentation {
-        FileRepresentation(contentType: .psbt) { file in
-            let url = FileManager.default.temporaryDirectory
-                .appendingPathComponent(file.filename)
-            try file.data.write(to: url)
-            return SentTransferredFile(url)
-        } importing: { url in
-            Log.warn("Importing PSBT files is not supported: \(url)")
-            // Importing closure - must return PSBTFile
-            guard let data = try? Data(contentsOf: url.file) else {
-                throw CocoaError(.fileReadUnknown)
-            }
-
-            return PSBTFile(data: data, filename: url.file.path())
-        }
-    }
-}
 
 extension UTType {
     static var psbt: UTType {
--- a/ios/Cove/ShareSheet.swift
+++ b/ios/Cove/ShareSheet.swift
@@ -108,6 +108,25 @@ enum ShareSheet {
         filename: String,
         completion: @escaping (Bool) -> Void
     ) {
+        guard let bytes = data.data(using: .utf8) else {
+            Log.error("Failed to encode share-sheet payload as UTF-8")
+            completion(false)
+            return
+        }
+        present(data: bytes, filename: filename, completion: completion)
+    }
+
+    /// Presents share sheet for binary data by writing to a temporary file
+    /// - Parameters:
+    ///   - data: the raw bytes to share
+    ///   - filename: the filename to use for the temporary file
+    ///   - completion: called after the share sheet dismisses with success/failure result
+    @MainActor
+    static func present(
+        data: Data,
+        filename: String,
+        completion: @escaping (Bool) -> Void
+    ) {
         guard let windowScene = UIApplication.shared.connectedScenes
             .compactMap({ $0 as? UIWindowScene })
             .first,
@@ -123,7 +142,7 @@ enum ShareSheet {
         let fileURL = tempDir.appendingPathComponent(filename)
 
         do {
-            try data.write(to: fileURL, atomically: true, encoding: .utf8)
+            try data.write(to: fileURL, options: .atomic)
         } catch {
             Log.error("Failed to write temp file for share sheet: \(error.localizedDescription)")
             completion(false)
--- a/ios/Cove/Flows/SendFlow/SendFlowHardwareScreen.swift
+++ b/ios/Cove/Flows/SendFlow/SendFlowHardwareScreen.swift
@@ -441,14 +441,18 @@ struct SendFlowHardwareScreen: View {
             app.nfcWriter.writeToTag(data: details.psbtBytes())
         }
 
-        ShareLink(
-            item: PSBTFile(data: details.psbtBytes(), filename: "transaction.psbt"),
-            preview: SharePreview(
-                "transaction.psbt - A Partially Signed Bitcoin Transaction",
-                image: Image(.bitcoinShield)
-            )
-        ) {
-            Text("More...")
+        Button("More...") {
+            // Defer presentation until the confirmationDialog's dismissal
+            // animation completes. A SwiftUI `ShareLink` presented directly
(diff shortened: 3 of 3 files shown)


pradhyum6144:
Opened #631 with a fix. Root cause is a presentation race: the SwiftUI `ShareLink` inside the export-transaction `.confirmationDialog` tries to present `UIActivityViewController` at the same moment the dialog's `UIAlertController` is dismissing, so Save to Files fails silently and share extensions like Signal surface a generic error. The fix routes "More..." through the existing `ShareSheet` helper (same one `MoreInfoPopover.exportTransactions` uses) with a 400ms delay so the dialog finishes dismissing first. Likely also closes #313.

ios/Cove/ShareSheet.swift
```diff
@@ -123,7 +142,7 @@ enum ShareSheet {
         let fileURL = tempDir.appendingPathComponent(filename)
 
         do {
-            try data.write(to: fileURL, atomically: true, encoding: .utf8)
+            try data.write(to: fileURL, options: .atomic)
```
praveenperera, line 145: whats this change?

ios/Cove/PSBTFile.swift
```diff
@@ -5,31 +5,8 @@
...
-            // Importing closure - must return PSBTFile
-            guard let data = try? Data(contentsOf: url.file) else {
-                throw CocoaError(.fileReadUnknown)
-            }
-
-            return PSBTFile(data: data, filename: url.file.path())
-        }
-    }
-}
-
```
praveenperera, line 32: is this removed since its no longer needed because we aren't using `ShareLink` anymore?

praveenperera:
this is good @pradhyum6144 thank you, where you able to build iOS and test it on a device?

pradhyum6144:
> this is good @pradhyum6144 thank you, where you able to build iOS and test it on a device?

I don't have a Mac to build-test this locally. Root cause analysis and the fix are in the PR description happy to iterate if anything looks off.  @praveenperera

ios/Cove/ShareSheet.swift
pradhyum6144: The old call data.write(to: fileURL, atomically: true, encoding: .utf8) is a String method  it tries to interpret the data as a UTF-8 string. PSBT files are binary data so using the string variant can corrupt the payload. The new call data.write(to: fileURL, options: .atomic) is the Data.write method, which writes the raw bytes directly. The .atomic option keeps the same safety guarantee (write-to-temp-then-rename) without the encoding conversion.

ios/Cove/PSBTFile.swift
pradhyum6144: Yes exactly. PSBTFile only existed to conform to Transferable for use with ShareLink. Since we're now using UIActivityViewController via the ShareSheet helper instead the Transferable conformance isn't needed anymore so this struct can be removed.

praveenperera:
@pradhyum6144 this is great thanks, I will test on device and merge

praveenperera (approved):

praveenperera:
@pradhyum6144 if you want you could explore if these other paths could also cause the same issue


  - Highest risk: ios/Cove/Flows/SendFlow/SendFlowHardwareScreen.swift:444. This ShareLink sits inside a confirmationDialog, which is a transient presenter and
    is the kind of place where share presentation tends to be fragile.
  - High risk: ios/Cove/Flows/SelectedWalletFlow/MoreInfoPopover.swift:82. This backup export is inside a toolbar Menu, via ios/Cove/Flows/SelectedWalletFlow/
    SelectedWalletScreen.swift:263. Same concern: nested transient presentation.
  - Lower risk: ios/Cove/Flows/TapSignerFlow/TapSignerSetupSuccessView.swift:77. This is a normal screen, so if the bug was specifically about dialog/menu
    presentation, this one is less likely to fail. If the bug was broader to ShareLink file export itself, it is still in scope.

pradhyum6144:
Thanks for pointing these out! I'm currently working on issue #146 once that's done, I'll tackle the highest risk one (SendFlowHardwareScreen.swift:444, the confirmationDialog case) today.

I'll have the remaining two (MoreInfoPopover.swift:82 and TapSignerSetupSuccessView.swift:77) done by tomorrow 6 PM UTC.

PR #632, opened 2026-04-15
Validate PSBT signatures at import time

Sandipmandal25:
### Validate PSBT signatures at import time

Unsigned PSBTs were passing through all import paths (hex, base64, raw bytes) fine, then blowing up later at finalization with a useless generic error. Not great UX when someone forgets to sign on their hardware wallet.

Added a `psbt_has_signatures` check right after parsing that covers all five BIP174 signature fields. If nothing’s signed, we return `NotSigned` immediately with a message that actually tells you what to do.

Also fixed `try_parse` which was using `if let Ok` and swallowing the error — it now propagates `NotSigned` specifically instead of silently falling through.

Note: adds a new `SignedImportError::NotSigned` variant so bindings need to be regenerated with `just build-ios` / `just build-android`.

## Summary by CodeRabbit

* **Bug Fixes**
  * PSBT imports now detect and reject unsigned PSBTs, returning a clear error instructing users to sign transactions (e.g., with a hardware wallet) before importing.

* **Tests**
  * Test suite updated with positive cases using explicitly signed PSBT bytes and new negative tests ensuring unsigned PSBTs are rejected.

### DIFF
--- a/rust/src/multi_format.rs
+++ b/rust/src/multi_format.rs
@@ -16,6 +16,7 @@ use tracing::{debug, warn};
 use crate::{
     hardware_export::HardwareExport,
     mnemonic::ParseMnemonic as _,
+    signed_import::SignedImportError,
     transaction::ffi::BitcoinTransaction,
     wallet::{AddressWithNetwork, address::AddressError},
 };
@@ -62,6 +63,9 @@ pub enum MultiFormatError {
 
     #[error("Taproot wallets are not supported yet")]
     TaprootNotSupported,
+
+    #[error("PSBT has no signatures — sign it with your hardware wallet before importing")]
+    PsbtNotSigned,
 }
 
 type Result<T, E = MultiFormatError> = std::result::Result<T, E>;
@@ -158,14 +162,21 @@ impl MultiFormat {
             }
         }
 
-        // try to parse as PSBT (base64 or hex)
-        if let Ok(parsed) = crate::signed_import::SignedTransactionOrPsbt::try_parse(string) {
-            if let Some(psbt) = parsed.psbt() {
-                return Ok(Self::SignedPsbt(psbt));
+        // try to parse as PSBT (base64 or hex); if it fails for any reason other than NotSigned,
+        // keep trying other formats — but NotSigned means it's definitely a PSBT so stop here
+        match crate::signed_import::SignedTransactionOrPsbt::try_parse(string) {
+            Ok(parsed) => {
+                if let Some(psbt) = parsed.psbt() {
+                    return Ok(Self::SignedPsbt(psbt));
+                }
+                if let Some(txn) = parsed.transaction() {
+                    return Ok(Self::Transaction(txn));
+                }
             }
-            if let Some(txn) = parsed.transaction() {
-                return Ok(Self::Transaction(txn));
+            Err(SignedImportError::NotSigned) => {
+                return Err(MultiFormatError::PsbtNotSigned);
             }
+            Err(_) => {}
         }
 
         warn!("could not parse string as MultiFormat: {string}");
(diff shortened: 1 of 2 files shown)


rust/src/multi_format.rs
```diff
@@ -158,14 +161,20 @@ impl MultiFormat {
             }
         }
 
-        // try to parse as PSBT (base64 or hex)
-        if let Ok(parsed) = crate::signed_import::SignedTransactionOrPsbt::try_parse(string) {
-            if let Some(psbt) = parsed.psbt() {
-                return Ok(Self::SignedPsbt(psbt));
+        // try to parse as PSBT (base64 or hex); propagate NotSigned so caller gets the clear error
```
praveenperera: need better explanation, we try to parse, if parsing fails for any reason other than NotSigned we try others, if its NotSigend that means it definitely is a PSBT so no point trying other formats

rust/src/multi_format.rs
```diff
@@ -158,14 +161,20 @@ impl MultiFormat {
...
+                if let Some(psbt) = parsed.psbt() {
+                    return Ok(Self::SignedPsbt(psbt));
+                }
+                if let Some(txn) = parsed.transaction() {
+                    return Ok(Self::Transaction(txn));
+                }
             }
-            if let Some(txn) = parsed.transaction() {
-                return Ok(Self::Transaction(txn));
+            Err(crate::signed_import::SignedImportError::NotSigned) => {
```
praveenperera: a use for this at the top of the file

praveenperera (commented):
Also maybe consider adding a new variant called `UnsignedPsbt<...>` and then the called can decide what to do with it, more flexible, but maybe not needed. Depends on who currently calls this function.

rust/src/multi_format.rs
Sandipmandal25: updated the comment!

rust/src/multi_format.rs
Sandipmandal25: moved to the top.

Sandipmandal25:
> Also maybe consider adding a new variant called `UnsignedPsbt<...>` and then the called can decide what to do with it, more flexible, but maybe not needed. Depends on who currently calls this function.

skipped for now since callers just surface the error message will add this if you think it's worth it !

praveenperera:
@Sandipmandal25 was the greptile issues resolved?


```
Two P1 gaps in multi_format.rs leave the UR and string-PSBT paths unguarded; the core signed_import logic is solid but the fix is incomplete.

The signed_import module's three paths are correctly guarded and the test suite is meaningfully improved. But MultiFormat::try_from_string still swallows NotSigned via if let Ok, and try_from_ur_payload never calls psbt_has_signatures at all — both are real import paths used in production (QR scanning, NFC). These two P1 issues should be addressed before merge.

rust/src/multi_format.rs — both the string PSBT branch (line 162) and the UR CryptoPsbt branch (lines 190–198) need signature validation.
```

Sandipmandal25:
> @Sandipmandal25 was the greptile issues resolved?
> 
> ```
> Two P1 gaps in multi_format.rs leave the UR and string-PSBT paths unguarded; the core signed_import logic is solid but the fix is incomplete.
> 
> The signed_import module's three paths are correctly guarded and the test suite is meaningfully improved. But MultiFormat::try_from_string still swallows NotSigned via if let Ok, and try_from_ur_payload never calls psbt_has_signatures at all — both are real import paths used in production (QR scanning, NFC). These two P1 issues should be addressed before merge.
> 
> rust/src/multi_format.rs — both the string PSBT branch (line 162) and the UR CryptoPsbt branch (lines 190–198) need signature validation.
> ```

yes, both fixed `try_from_string `now uses match to propagate `NotSigned` as `PsbtNotSigned`, and `try_from_ur_payload `has the `psbt_has_signatures `check added. also added regression tests for both paths.

praveenperera:
Thank you @Sandipmandal25, merged!

PR #633, opened 2026-04-15
Improve NFC reliability for TAPSIGNER backup

### LINKED ISSUE #623: Tapsigner backup keeps failing
I can't get the tapsigner backup to work. Using grapheneos.

pradhyum6144:
Fixes #623

TapSigner backup fails on GrapheneOS because the default IsoDep timeout (~2s) is too short for the multi-step backup APDU exchange.

**Changes:**
- Increase IsoDep timeout to 5s (8s for backup operations)
- Add retry with reconnect for recoverable NFC errors (TagLostException, IOException)
- Differentiate between recoverable and non-recoverable NFC errors
- Show proactive UX guidance during backup: "Keep your phone steady on the card"
- Improve error message when connection is lost

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NFC timeout handling with longer timeout for backup flows and better recovery when connections are lost
  * More helpful error messages on NFC failures, now prompting users to hold still and try again

* **New Features**
  * On-screen guidance during backup operations instructing users to keep the phone steady for the longer backup duration

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/nfc/TapCardNfcManager.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/nfc/TapCardNfcManager.kt
@@ -120,15 +120,31 @@ class TapCardNfcManager private constructor() {
                                 IsoDep.get(detectedTag)
                                     ?: throw Exception("Tag doesn't support IsoDep (ISO7816)")
 
-                            // connect to tag
+                            // connect to tag with increased timeout for reliability
                             if (!isoDep.isConnected) {
                                 isoDep.connect()
                             }
 
-                            Log.d(tag, "Connected to IsoDep tag")
+                            // use higher timeout for backup (heavier NFC operation)
+                            val isBackup = cmd is TapSignerCmd.Backup
+                            val timeout = if (isBackup) {
+                                TapCardTransport.ISODEP_BACKUP_TIMEOUT_MS
+                            } else {
+                                TapCardTransport.ISODEP_TIMEOUT_MS
+                            }
+                            isoDep.timeout = timeout
 
-                            // create transport with message callback
-                            val transport = TapCardTransport(isoDep, onMessageUpdate)
+                            Log.d(tag, "Connected to IsoDep tag (timeout=${timeout}ms, backup=$isBackup)")
+
+                            // send proactive UX guidance for backup operations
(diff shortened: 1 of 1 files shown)


praveenperera:
thanks @pradhyum6144 we you able to test this on a device? Also reply to or address the greptile comments.

pradhyum6144:
@praveenperera  Wasn't able to test on device because fresh debug builds crash on launch (cloud storage is not initialized in onBootstrapComplete)  filed that as #634. The code compiles cleanly though, and the change is minimal (33 insertions, 5 deletions in one file).

Let me know if there's anything else! and also Greptile comments addressed

praveenperera (changes requested):
Apply backup timeout to TapSigner setup flows `android/app/src/main/java/org/bitcoinppl/cove/nfc/TapCardNfcManager.kt:129-133` TapSignerCmd.Setup also executes the same backup() APDU sequence during initial setup and continueSetup (the Rust reader's setup path calls `backup_change_xpub before derive/change`), but this branch only enables `ISODEP_BACKUP_TIMEOUT_MS` for `TapSignerCmd.Backup`. 

That means creating a new TapSigner or resuming setup still uses the shorter 5s timeout here, so the timeout this patch is trying to avoid can still happen in those flows even though direct backup now succeeds.

pradhyum6144:
> Apply backup timeout to TapSigner setup flows `android/app/src/main/java/org/bitcoinppl/cove/nfc/TapCardNfcManager.kt:129-133` TapSignerCmd.Setup also executes the same backup() APDU sequence during initial setup and continueSetup (the Rust reader's setup path calls `backup_change_xpub before derive/change`), but this branch only enables `ISODEP_BACKUP_TIMEOUT_MS` for `TapSignerCmd.Backup`.
> 
> That means creating a new TapSigner or resuming setup still uses the shorter 5s timeout here, so the timeout this patch is trying to avoid can still happen in those flows even though direct backup now succeeds.

Applied the backup timeout to Setup flows as well  not just TapSignerCmd.Backup, but also TapSignerCmd.Setup (and continueSetup), since the Rust reader runs the same backup APDUs during initial setup. Both flows now use the 8s timeout.

praveenperera (approved):

praveenperera:
thanks @pradhyum6144 merged

PR #635, opened 2026-04-15
feat: tapsigner pin length increase

### LINKED ISSUE #488: Support custom length PINs for TAPSIGNER (not just 6 digits)
Currently, TAPSIGNER PIN input is restricted to exactly 6 digits. TAPSIGNER devices actually support PINs of varying lengths (minimum 6, maximum 32 characters).

Keep 6 as the default, but have an "Advanced PIN" sheet that is similar looking but with a text input box instead of 6 boxes.

Image

oyetanishq:
close: #488 

- Accept 6-32 character PINs (only for ios)

image

## Summary by CodeRabbit

* **New Features**
  * PIN input now supports 6–32 characters (previously limited to 6)
  * Added character count display to PIN entry screens
  * Introduced explicit "Continue" buttons for PIN submission

* **Bug Fixes**
  * Fixed typo in PIN length validation error message
  * Corrected PIN validation logic to accept minimum 6-character PINs

### DIFF
--- a/ios/Cove/Flows/TapSignerFlow/TapSignerConfirmPinView.swift
+++ b/ios/Cove/Flows/TapSignerFlow/TapSignerConfirmPinView.swift
@@ -123,13 +123,16 @@ struct TapSignerConfirmPinView: View {
                 }
                 .padding(.horizontal)
 
-                HStack {
-                    ForEach(0 ..< 6, id: \.self) { index in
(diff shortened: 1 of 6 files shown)


praveenperera:
No this is the wrong approach, we want to keep 6 digits as the default, but have an `Advanced PIN` screen that will just be a textbox so you can enter whatever you want.

Pin4

PR #643, opened 2026-04-16
Android: add visual affordance to amount input

### LINKED ISSUE #515: Android: Add visual affordance to amount input field
## Problem
Users report it's unclear you can tap to manually type send amount. The amount TextField lacks visual feedback.

## Current Behavior
`EnterAmountView.kt` TextField has no `.clickable()` modifier - no ripple effect when touched.

## Proposed Solution
Add `.clickable` modifier to amount field for visual feedback, or add label "Tap to enter amount".

## Files
- `android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/EnterAmountView.kt:102-132`

## iOS Status
Not an issue - iOS TextField conventions are inherently clear.

pradhyum6144:
Combines the three directions from the design discussion on #515. The amount field on Android is a BasicTextField with no Material chrome, so users had no visual cue that it was tappable.

Adds a "Tap to enter amount" placeholder centered in the Box when the value is empty or "0" and the field isn't focused. The underlying text is drawn transparent in that state so the default "0" doesn't bleed through behind the placeholder. Wraps the amount Box in .clickable { focusRequester.requestFocus() } so taps on the empty area around the glyphs give a Material ripple and focus the field BasicTextField still consumes taps within its own bounds, so the wrapper only fires for the outer area. Drops a HorizontalDivider under the amount row as a subtle bottom border, matching the conventional "this is an input" look.

Falls back to a locally-remembered FocusRequester when the caller doesn't provide one, so the clickable wrapper always has a target.

Closes #515

## Release Notes

* **Bug Fixes**
  * Enhanced the amount input field in the Send flow with improved interaction responsiveness and expanded tap targets for easier focus activation.
  * Added placeholder guidance text when the input field is empty.
  * Refined visual layout with a divider separating the input area from controls below.

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/EnterAmountView.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/EnterAmountView.kt
@@ -17,6 +17,7 @@ import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.ArrowDropDown
 import androidx.compose.material3.DropdownMenu
 import androidx.compose.material3.DropdownMenuItem
+import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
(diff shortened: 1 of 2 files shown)


praveenperera:
looks good @pradhyum6144 can you include a video showing what it looks like now, thanks

pradhyum6144:
> looks good @pradhyum6144 can you include a video showing what it looks like now, thanks

Hey @praveenperera, I tried running the app on my device to record the video but it crashes on startup with cloud storage is not initialized looks like it needs Google cloud credentials configured which I don't have set up locally. Is there a way to disable cloud storage for local dev builds? Also if possible can you test on your device.??

praveenperera:
@pradhyum6144 ill fix this and let you know, whats the error you get?

praveenperera:
@coderabbitai review
@greptileai

PR #644, opened 2026-04-16
detect silent payment addresses and surface a clear error

### LINKED ISSUE #617: Add support for sending to silent payment addresses
Track support for sending to silent payment addresses in Cove

guptapratykshh:
Fixes #617

blocks sending to silent payment addresses with a clear "not yet supported" message instead of a confusing "Invalid Address" error.

## Summary by CodeRabbit

* **New Features**
  * Detects and rejects unsupported silent-payment addresses (sp1.../tsp1...) when entered, scanned, or embedded in URs/QRs.
  * Shows a clear "Unsupported Address" alert explaining silent-payment addresses aren’t supported yet, then dismisses the alert and focuses the address field.

* **Bug Fixes**
  * QR scanner and camera error handling present an explicit "Unsupported Address" message for silent-payment scans instead of a generic error.

### DIFF
--- a/rust/crates/cove-types/src/address.rs
+++ b/rust/crates/cove-types/src/address.rs
@@ -98,6 +98,9 @@ pub enum AddressError {
 
     #[error("empty address")]
     EmptyAddress,
+
+    #[error("silent payment addresses (sp1...) are not yet supported for sending")]
+    SilentPaymentNotSupported,
 }
 
 impl Clone for AddressInfo {
@@ -208,6 +211,28 @@ impl AddressWithNetwork {
     }
 }
 
+fn is_silent_payment_address(s: &str) -> bool {
+    use bitcoin::bech32::{Bech32m, primitives::decode::CheckedHrpstring};
+
+    let s = s.trim();
+
+    // Quick prefix check before attempting a full decode.
+    let lower = s.to_lowercase();
+    if !lower.starts_with("sp1") && !lower.starts_with("tsp1") {
+        return false;
+    }
+
+    // BIP352 silent payment addresses are bech32m-specifically (not bech32).
+    // Reject strings that merely look like SP addresses but fail the checksum,
+    // and reject strings with a valid bech32 (not bech32m) checksum.
+    let Ok(checked) = CheckedHrpstring::new::<Bech32m>(s) else {
+        return false;
+    };
+
+    let hrp = checked.hrp().to_lowercase();
+    hrp == "sp" || hrp == "tsp"
+}
+
 fn parse_bitcoin_uri(input: &str) -> Result<ParsedBitcoinUri, Error> {
     let input = input.trim();
     if input.is_empty() {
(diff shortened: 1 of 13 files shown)


rust/crates/cove-types/src/address.rs
```diff
@@ -65,7 +65,7 @@ pub struct AddressWithNetwork {
 type Error = AddressError;
 type Result<T, E = Error> = std::result::Result<T, E>;
 
-#[derive(Debug, PartialEq, Eq, thiserror::Error, uniffi::Error)]
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error, uniffi::Error)]
```
praveenperera: why did we need `Clone` here?

rust/src/manager/send_flow_manager/state.rs
```diff
@@ -44,6 +45,7 @@ pub struct SendFlowManagerState {
     pub max_selected: Option<Arc<Amount>>,
 
     pub address: Option<Arc<Address>>,
+    pub(crate) address_error: Option<SendFlowError>,
```
praveenperera: do we need a new field here? maybe just use the existing alert mechanism to show an alert with the error.

praveenperera (changes requested):
also read CONTRIBUTING.md and follow the commit message format

praveenperera (changes requested):
- Regenerate UniFFI bindings for the new send-flow error, read `CONTRIBUTING.md`
  - Use the URI-aware parser when surfacing silent-payment alerts `send_flow_manager.rs:426-428` If the user pastes a BIP21 URI such as bitcoin:sp1... into the send field and then submits, this path reparses with Address::from_string, which only accepts raw addresses. In that scenario the new SilentPaymentNotSupported branch is never reached and the app still shows the generic invalid-address alert, even though AddressWithNetwork::try_new already recognizes silent-payment URIs and has tests for them.

praveenperera (changes requested):
@guptapratykshh  Validate silent-payment candidates before surfacing unsupported `crates/cove-types/src/address.rs`

Any string that merely starts with `sp1q` or `tsp1q` is currently classified as `SilentPaymentNotSupported`, even if it is truncated, has a bad checksum, or is otherwise not a real silent-payment address. In those cases, users will see the new unsupported-address path instead of the existing invalid-address path, which is a user-visible validation regression. Please only return `SilentPaymentNotSupported` after confirming the input is a valid silent-payment encoding.

praveenperera (commented):
rust/src/multi_format.rs:113-120 does not handle the new AddressError::SilentPaymentNotSupported. The shared scanner still goes through MultiFormat::try_from_string() (rust/src/qr_scanner.rs:218, :276, :399), so an sp1... QR now fails as a generic parse error before Send Flow reaches handle_scan_code_changed() (rust/src/manager/send_flow_manager.rs:1493-1497). On iOS that lands in the generic QR error path in ios/Cove/Flows/SendFlow/SetAmountScreen/QrCodeAddressView.swift:99-111, and on Android Send Flow only accepts MultiFormat. Address in android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt:289-307.  The new unsupported-address copy therefore works for manual entry, but not for scanned silent-payment addresses. 

Please thread the new error through the shared scanner path, or otherwise let Send Flow see the raw scanned string.rust/crates/cove-types/src/address.rs:217-225 uses bitcoin::bech32::decode() as the validator. 

That API accepts either Bech32 or Bech32m checksums, so a checksummed-but-non-Bech32m sp1.../tsp1... string will be reported as SilentPaymentNotSupported instead of InvalidAddress. Silent payment detection needs a Bech32m-only checksum check here.I

praveenperera (changes requested):

rust/src/qr_scanner.rs
```diff
@@ -395,8 +405,7 @@ impl QrScanner {
         }
 
         // plain string - address, xpub, etc.
-        let multi_format =
-            MultiFormat::try_from_string(&qr).map_err_str(MultiQrError::ParseError)?;
+        let multi_format = MultiFormat::try_from_string(&qr).map_err(map_multi_format_err)?;
```
praveenperera: once you do the above you can just do `?` or if needed only `let multi_format = MultiFormat::try_from_string(qr).map_err(MultiQrError::from)?;`

praveenperera (changes requested):
looks good, but need some small changes

Issue #314, opened 2025-08-24
Validation error on amount field paste


praveenperera:
Looks like there is an issue with input type validation on send flows. It encoded the bitcoin address into a number and puts it in as a send value

praveenperera:
Closed by PR #645

PR #645, opened 2026-04-16
Fix fiat amount paste stripping non-numeric characters (#314)

pradhyum6144:
Pasting a Bitcoin address into the send-flow amount field while in fiat mode caused the address to be silently stripped down to just its digits and treated as a send value, as described in #314. The fiat `on_change` handler filtered characters through `c.is_numeric() || c == '.'`, so input like `bc1qxyz0123456789abc` collapsed to `0123456789` and was parsed as a fiat amount. The BTC handler was already safe because it uses strict `parse::<u64>` / `parse::<f64>` and falls back to the previous value on failure.

This change rejects any input containing an alphabetic character at the top of the fiat handler, returning the previous value unchanged so the text field does not accept addresses or other alphanumeric pastes. Tests cover bech32 addresses, legacy P2PKH addresses, a single stray letter, plain digits, decimals, and empty input.

## Summary by CodeRabbit

* **New Features**
  * Added paste sanitization for fiat and BTC amounts; common currency/unit tokens are stripped and public sanitization utilities added.

* **Bug Fixes**
  * Cleaned pasted values are preserved instead of being lost by early-exit logic.
  * Malformed pastes (addresses, lone letters, unsupported tokens, or mixed currency tokens) now revert to the previous valid value.

* **Tests**
  * Added tests covering accepted/rejected paste cases, suffix stripping, edge inputs, and BTC/fiat mixing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bitcoinppl/cove/pull/645?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/rust/src/manager/send_flow_manager/btc_on_change.rs
+++ b/rust/src/manager/send_flow_manager/btc_on_change.rs
@@ -43,15 +43,23 @@ impl BtcOnChangeHandler {
         }
 
         let old = old_value.trim();
+        let new_value_trimmed = new_value.trim();
 
         // strip BTC-unit tokens from pasted input (e.g. "100 SATS" → "100"). #314
-        let Some(new) = sanitize::sanitize_btc_amount(new_value.trim()) else {
+        let Some(new) = sanitize::sanitize_btc_amount(new_value_trimmed) else {
             return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() };
         };
 
+        // If sanitization stripped tokens (e.g. "100 BTC" → "100"), early exits must still
+        // emit the cleaned string so the raw pasted text doesn't stay visible in the UI.
+        let sanitization_changed = new != new_value_trimmed;
+
         // early exit if nothing changed
         if old == new {
-            return Changeset::default();
+            return Changeset {
+                entering_amount_btc: sanitization_changed.then_some(old.into()),
+                ..Default::default()
+            };
         }
 
         if new == "00" {
@@ -102,7 +110,10 @@ impl BtcOnChangeHandler {
 
         // if the unformatted is the same as the old value, then we don't need to do anything
         if old.replace(',', "") == unformatted {
-            return Changeset::default();
+            return Changeset {
+                entering_amount_btc: sanitization_changed.then_some(old.into()),
+                ..Default::default()
+            };
         }
 
         // ---------------------------------------------------------------------
--- a/rust/src/manager/send_flow_manager/btc_on_change.rs
+++ b/rust/src/manager/send_flow_manager/btc_on_change.rs
@@ -47,8 +47,13 @@ impl BtcOnChangeHandler {
         // strip currency tokens from pasted input (e.g. "100 SATS" → "100"). #314
         let sanitized;
         let new = match sanitize::sanitize_amount(new_value.trim()) {
-            Some(s) => { sanitized = s; sanitized.as_str() }
-            None => return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() },
+            Some(s) => {
+                sanitized = s;
+                sanitized.as_str()
+            }
+            None => {
+                return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() };
+            }
         };
 
         // early exit if nothing changed
--- a/rust/src/manager/send_flow_manager/btc_on_change.rs
+++ b/rust/src/manager/send_flow_manager/btc_on_change.rs
@@ -44,8 +44,8 @@ impl BtcOnChangeHandler {
 
         let old = old_value.trim();
 
-        // strip currency tokens from pasted input (e.g. "100 SATS" → "100"). #314
-        let Some(new) = sanitize::sanitize_amount(new_value.trim()) else {
+        // strip BTC-unit tokens from pasted input (e.g. "100 SATS" → "100"). #314
+        let Some(new) = sanitize::sanitize_btc_amount(new_value.trim()) else {
             return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() };
         };
 
--- a/rust/src/manager/send_flow_manager/btc_on_change.rs
+++ b/rust/src/manager/send_flow_manager/btc_on_change.rs
@@ -45,15 +45,8 @@ impl BtcOnChangeHandler {
         let old = old_value.trim();
 
         // strip currency tokens from pasted input (e.g. "100 SATS" → "100"). #314
-        let sanitized;
-        let new = match sanitize::sanitize_amount(new_value.trim()) {
-            Some(s) => {
-                sanitized = s;
-                sanitized.as_str()
-            }
-            None => {
-                return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() };
-            }
+        let Some(new) = sanitize::sanitize_amount(new_value.trim()) else {
+            return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() };
         };
 
         // early exit if nothing changed
--- a/rust/src/manager/send_flow_manager/btc_on_change.rs
+++ b/rust/src/manager/send_flow_manager/btc_on_change.rs
@@ -5,6 +5,7 @@ use cove_types::{amount::Amount, unit::BitcoinUnit};
 use cove_util::format::{self, NumberFormatter as _};
 use tracing::debug;
 
+use super::sanitize;
 use super::state::State;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


praveenperera:
i'm not sure if this is an issue, this would also break people copying in amounts that accidentally has letters in it, would this break people copying with $ or other fiat prefixes or BTC or SATS suffixes? @pradhyum6144

pradhyum6144:
Good catch $/€ prefixes are fine since they're not alphabetic, but you're right that 50 BTC or 100 SATS would break with this approach.

pradhyum6144:
A better fix would be to strip known currency suffixes/prefixes (BTC, SATS, sat, USD, EUR, etc.) before the alphabetic check if stripping those leaves a valid number, accept it; only reject if alphabetic characters remain after that cleanup. That way 100 SATS → 100 works, but a full bech32 address still gets rejected.

Want me to revise with that approach?

praveenperera:
I think we need to make it work for CHF too, but overall i'm not sure if this change is needed. Can you make a case for why its important?

pradhyum6144:
> I think we need to make it work for CHF too, but overall i'm not sure if this change is needed. Can you make a case for why its important?

without this fix, pasting a Bitcoin address into the amount field silently mangles it into a number (e.g. bc1q... becomes some large digit sequence) and could be submitted as a send amount. In a Bitcoin wallet that's a meaningful risk  the user thinks they copied an amount, the app silently accepts something else.

On CHF: the revised approach handles it. Instead of blanket-rejecting any input with letters, we'd strip known currency suffixes/prefixes (BTC, SATS, sat, CHF, USD, EUR, etc.) first. If what remains is a valid number, accept it 100 CHF → 100. Only reject if alphabetic characters survive that cleanup, which a bech32 address always will.

Happy to implement that if you want to move forward with it.

praveenperera:
Okay I think thats fair go ahead @pradhyum6144 thanks

pradhyum6144:
> Okay I think thats fair go ahead @pradhyum6144 thanks

sure sir i will start working on this after completing #653 almost done with that enhancement.

pradhyum6144:
Done! Here's what I implemented:

**Approach:** Instead of blanket-rejecting any input with letters, we now strip known currency tokens first (BTC, SAT, SATS, CHF, USD, EUR, GBP, CAD, AUD, JPY  case-insensitive, prefix or suffix), then only reject if alphabetic characters survive that cleanup.

- `100 CHF` → `100`
- `BTC 0.5` → `0.5` 
- `1000 SATS` → `1000` 
- `bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq` → rejected (letters survive after stripping)

The same logic applies to both the fiat and BTC amount fields. Added 13 unit tests across the new `sanitize::strip_currency_suffix` helper and both on-change handlers.

rust/src/manager/send_flow_manager/btc_on_change.rs
```diff
@@ -43,7 +44,29 @@ impl BtcOnChangeHandler {
...
+                        return Changeset {
+                            entering_amount_btc: Some(old.to_string()),
+                            ..Default::default()
+                        };
+                    }
+                }
+            } else {
+                raw
+            }
+        };
```
praveenperera, line 51: move this to sanitize since this is a sanitization, and use early returns to reduce nesting

rust/src/manager/send_flow_manager/fiat_on_change.rs
```diff
@@ -65,6 +65,27 @@ impl FiatOnChangeHandler {
...
+                None => {
+                    return Ok(Changeset {
+                        entering_fiat_amount: Some(old_value.to_string()),
+                        ..Default::default()
+                    });
+                }
+            }
+        } else {
+            new_value
+        };
```
praveenperera: seems kinda of like duplicated logic to the one above? can we we combine into one santize function?

rust/src/manager/send_flow_manager/sanitize.rs
```diff
@@ -1,3 +1,41 @@
...
+            }
+        }
+    }
+
+    if work.chars().any(|c| c.is_alphabetic()) {
+        return None;
+    }
+
+    Some(work.to_string())
+}
```
praveenperera, line 49: this seemes more complicated than it needs to be, try something like this

```rust
fn strip_prefix_ignore_ascii_case<'a>(input: &'a str, token: &str) -> Option<&'a str> {
    input
        .get(..token.len())
        .filter(|prefix| prefix.eq_ignore_ascii_case(token))
        .map(|_| input[token.len()..].trim())
}

fn strip_suffix_ignore_ascii_case<'a>(input: &'a str, token: &str) -> Option<&'a str> {
    input
        .get(input.len().checked_sub(token.len())?..)
        .filter(|suffix| suffix.eq_ignore_ascii_case(token))
        .map(|_| input[..input.len() - token.len()].trim())
}

pub fn strip_currency_suffix(input: &str) -> Option<String> {
    let mut work = input.trim();

    loop {
        let next = CURRENCY_TOKENS.iter().find_map(|token| {
            strip_prefix_ignore_ascii_case(work, token)
                .or_else(|| strip_suffix_ignore_ascii_case(work, token))
        });

        match next {
            Some(stripped) => work = stripped,
            None => break,
        }
    }

    if work.chars().any(|c| c.is_alphabetic()) {
        return None;
    }

    Some(work.to_string())
}
```

praveenperera (changes requested):

rust/src/manager/send_flow_manager/fiat_on_change.rs
```diff
@@ -63,6 +63,21 @@ impl FiatOnChangeHandler {
         let old_value = old_value.trim();
         let new_value = new_value.trim();
 
+        // strip currency tokens from pasted amounts (e.g. "$12.50", "12.50 USD")
+        let sanitized_fiat;
```
praveenperera: why do we need this var?

praveenperera (changes requested):

rust/src/manager/send_flow_manager/fiat_on_change.rs
pradhyum6144: sir It's a Rust lifetime requirement sanitize_amount returns Option<String> and we need a &str that stays alive for the rest of the function. The String has to be owned somewhere; sanitized_fiat is that storage so sanitized_fiat.as_str() remains valid. Without it the String would drop at the end of the match arm and the borrow wouldn't compile.

praveenperera: keep it simple

```rust
        let new = match sanitize::sanitize_amount(new_value.trim()) {
            Some(s) => s,
            None => {
                return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() };
            }
        };

```

praveenperera: or you could do

```rust
        let Some(new) = sanitize::sanitize_amount(new_value.trim()) else {
            return Changeset { entering_amount_btc: Some(old.into()), ..Default::default() };
        };
```

praveenperera:
@pradhyum6144 please address the AI bot comments

rust/src/manager/send_flow_manager/sanitize.rs
```diff
@@ -1,9 +1,77 @@
+const FIAT_TOKENS: &[&str] = &[
+    // fiat symbols
+    "$", "€", "£", "¥", "₹", // fiat codes
+    "USD", "EUR", "GBP", "JPY", "CAD", "AUD", "CHF", "INR", "MXN",
+];
```
praveenperera: can we get these from the fiat currency type we already have, my concern is if we had new fiat currencies in the future we might forget to update this list here.

praveenperera (changes requested):

rust/src/manager/send_flow_manager/sanitize.rs
pradhyum6144: ok sir soo i Replaced the hardcoded FIAT_TOKENS list with a fiat_tokens() function that derives directly from FiatCurrency:

`fn fiat_tokens() -> Vec<&'static str> {
    let mut tokens: Vec<&'static str> = FiatCurrency::all_symbols().to_vec();
    for currency in FiatCurrency::iter() {
        tokens.push(currency.into());
    }
    tokens
} `
Now if a new currency is added to the FiatCurrency enum in fiat.rs, its symbol and code automatically appear in the sanitizer no separate list to maintain and All 8 tests still pass.

praveenperera (changes requested):
`rust/src/manager/send_flow_manager/btc_on_change.rs:47` and `rust/src/manager/send_flow_manager/fiat_on_change.rs:67` sanitize first, but the later early exits (old == new / old_value_raw == new_value_raw) still return a default changeset when the sanitized text collapses to the existing numeric value. Because sanitizeBtcEnteringAmount() / sanitizeFiatEnteringAmount() only rewrite the field when entering_* is set, pastes like 100 BTC over 100, or USD 100 / 100 CHF over $100, will stay unstripped in the UI. This patch needs to emit the sanitized display string whenever sanitization changed the input, even if the parsed amount did not change

pradhyum6144:
The issue was that early exits (old == new / old_value_raw == new_value_raw) returned Changeset::default()  no entering_* set  so the UI never rewrote the field even though sanitization had already stripped the tokens.
Both handlers now track sanitization_changed = sanitized != original_input. When that's true, every early exit emits the existing display string as entering_fiat_amount / entering_amount_btc, forcing the field to rewrite and clear the pasted tokens.

praveenperera:
Closes: #314

praveenperera:
thanks @pradhyum6144 merged

PR #646, opened 2026-04-16
Unify bitcoin: URI handling between text input and QR scan

### LINKED ISSUE #647: Pasting bitcoin: URI into address field fails but scanning works
Pasting a `bitcoin:` URI (e.g. `bitcoin:bc1qxxx?amount=0.001`) into the send-flow address text field silently fails,  the address field stays empty. Scanning the exact same URI as a QR code works correctly and also fills in the amount.

The text input path uses `Address::from_string`, which only accepts raw addresses, while the QR scan path uses `AddressWithNetwork::try_new`, which handles the full `bitcoin:` URI format. Both paths should behave the same.

Sandipmandal25:
Pasting a `bitcoin:` URI into the send-flow address field was silently failing, while scanning the exact same URI as a QR code worked fine.

The two paths were using different parsers, text input called `Address::from_string` (which only accepts raw addresses), while the QR scan path used `AddressWithNetwork::try_new` (which supports full `bitcoin:` URIs including scheme and query params).

This change unifies both paths to use the same parser, so pasting `bitcoin:bc1qxxx?amount=0.001` now correctly sets both the address and the amount, matching the behaviour from QR scanning.
 fixes #647 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved address handling to recognize and normalize bitcoin: URIs, automatically cleaning and updating the input field so pasted URIs display as canonical addresses.
  * When amounts are embedded in address URIs, the app now extracts and applies the amount (and clears max-selection when appropriate), ensuring payment amounts are populated correctly.

### DIFF
--- a/rust/src/manager/send_flow_manager.rs
+++ b/rust/src/manager/send_flow_manager.rs
@@ -840,17 +840,42 @@ impl RustSendFlowManager {
         Some(())
     }
 
+    /// Called when the user types or pastes into the address field.
+    /// Handles plain addresses and full bitcoin: URIs (extracts amount if present).
     fn handle_entering_address_changed(self: &Arc<Self>, address: String) {
         debug!("handle_entering_address_changed: {address}");
 
         let mut sender = DeferredSender::new(self.reconciler.clone());
 
-        // update the state
         self.state.lock().entering_address = address.clone();
 
-        // if the address is valid, then set it in the state
-        let address = Address::from_string(&address, self.state.lock().metadata.network).ok();
-        let address = address.map(Arc::new);
+        let network = self.state.lock().metadata.network;
+        let parsed = AddressWithNetwork::try_new(&address).ok();
+        let parsed = parsed
+            .filter(|address_with_network| address_with_network.is_valid_for_network(network));
+
+        // if input was a URI, show just the address in the text field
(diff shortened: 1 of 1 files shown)


rust/src/manager/send_flow_manager.rs
```diff
@@ -840,17 +840,40 @@ impl RustSendFlowManager {
...
 
-        // if the address is valid, then set it in the state
-        let address = Address::from_string(&address, self.state.lock().metadata.network).ok();
-        let address = address.map(Arc::new);
+        let network = self.state.lock().metadata.network;
+        let parsed = AddressWithNetwork::try_new(&address).ok();
+        let parsed = parsed.filter(|awn| awn.is_valid_for_network(network));
+
+        // if input was a URI, show just the address in the text field
+        if let Some(awn) = &parsed {
```
praveenperera: whats `awn` use longer variable names

praveenperera (changes requested):

rust/src/manager/send_flow_manager.rs
Sandipmandal25: updated with address_with_network

praveenperera:
Thanks @Sandipmandal25 approved

praveenperera (approved):

PR #648, opened 2026-04-16
let users download their TAPSIGNER backup when setup fails partway

### LINKED ISSUE #308: TAPSIGNER intermediate screens
If it fails after backup, don't show are error, give download and continue, etc....

guptapratykshh:
Fixes #308.

When TAPSIGNER setup fails after the backup step has succeeded, show an intermediate screen with a download backup button instead of a generic error so users do not lose the backup the card already created.

## Release Notes

* **New Features**
  * Backup download option now available during Tap Signer setup recovery when backup data is present
  * New recovery status screen guides users when backup information is available

* **Bug Fixes**
  * Improved setup recovery process with enhanced retry and continuation logic

### DIFF
--- a/ios/Cove/Flows/TapSignerFlow/TapSignerSetupRetryView.swift
+++ b/ios/Cove/Flows/TapSignerFlow/TapSignerSetupRetryView.swift
@@ -15,21 +15,28 @@ struct TapSignerSetupRetry: View {
     let tapSigner: TapSigner
     let response: SetupCmdResponse
 
-    var body: some View {
-        VStack(spacing: 40) {
-            VStack {
-                HStack {
-                    Button(action: { manager.popRoute() }) {
-                        Text("Cancel")
-                    }
+    @State private var isRunning = false
+    @State private var didSaveBackup = false
 
-                    Spacer()
-                }
-                .padding(.top, 20)
-                .padding(.horizontal, 10)
-                .foregroundStyle(.primary)
-                .fontWeight(.semibold)
-            }
+    private var availableBackup: Data? {
+        switch response {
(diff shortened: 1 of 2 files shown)


ios/Cove/Flows/TapSignerFlow/TapSignerSetupRetryView.swift
```diff
@@ -15,21 +15,32 @@ struct TapSignerSetupRetry: View {
     let tapSigner: TapSigner
     let response: SetupCmdResponse
 
+    /// Backup bytes are available after the backup step has succeeded, even if
+    /// a later step (derive / change pin) failed. Issue #308: surface the
```
praveenperera: remove reference to issue, keep comments that make sense without issue context

ios/Cove/Flows/TapSignerFlow/TapSignerSetupRetryView.swift
```diff
@@ -56,48 +67,152 @@ struct TapSignerSetupRetry: View {
...
+        }
+        .background(patternBackground)
+        .scrollIndicators(.hidden)
+        .navigationBarHidden(true)
+    }
+
+    /// Shown when backup succeeded but a later step failed. The user can
+    /// download the backup now and continue to finish the remaining steps.
+    @ViewBuilder
+    func intermediateBody(backup: Data) -> some View {
```
praveenperera: is there a better name we can use instead of intermediateBoy?

praveenperera (changes requested):

praveenperera (changes requested):
can we bring this change over to android too?

PR #649, opened 2026-04-16
[Feat] Custom block explorer

### LINKED ISSUE #544: Feature Request: Add custom block explorer option
I would like to be able to add my own custom block explorer when clicking on an on-chain transaction ID. (Android)

Vineet1101:
Fixes #544

## Summary by CodeRabbit

* **New Features**
  * Configure custom block explorer URLs for Bitcoin, Testnet, Testnet4, and Signet.
  * Transaction links now prefer configured explorers, with automatic fallback to defaults.
  * Custom URLs may include placeholders for transaction IDs or be treated as base paths.

* **Bug Fixes / Improvements**
  * Empty URL clears the custom explorer; non-HTTP(S) URLs are rejected.

### DIFF
--- a/rust/src/database/global_config.rs
+++ b/rust/src/database/global_config.rs
@@ -316,14 +316,14 @@ impl GlobalConfigTable {
     }
 
     #[uniffi::method(name = "customBlockExplorer")]
-    pub fn _custom_block_explorer(&self, network: Network) -> Option<String> {
+    pub fn get_custom_block_explorer(&self, network: Network) -> Option<String> {
         self.custom_block_explorer(network)
     }
 
     #[uniffi::method(name = "setCustomBlockExplorer")]
-    pub fn _set_custom_block_explorer(&self, network: Network, url: String) -> Result<()> {
+    pub fn set_custom_block_explorer(&self, network: Network, url: String) -> Result<()> {
         if url.is_empty() {
-            return self._clear_custom_block_explorer(network);
+            return self.clear_custom_block_explorer(network);
         }
 
         if !url.starts_with("https://") && !url.starts_with("http://") {
@@ -337,7 +337,7 @@ impl GlobalConfigTable {
     }
 
     #[uniffi::method(name = "clearCustomBlockExplorer")]
-    pub fn _clear_custom_block_explorer(&self, network: Network) -> Result<()> {
+    pub fn clear_custom_block_explorer(&self, network: Network) -> Result<()> {
         let key = GlobalConfigKey::CustomBlockExplorer(network);
         self.delete(key)
     }
--- a/rust/src/database/global_config.rs
+++ b/rust/src/database/global_config.rs
@@ -304,6 +304,25 @@ impl GlobalConfigTable {
         self.set_priv_hashed_pin_code(hashed_pin_code)
     }
 
+    #[uniffi::method(name = "customBlockExplorer")]
+    pub fn _custom_block_explorer(&self) -> Option<String> {
+        self.custom_block_explorer().ok().filter(|s| !s.is_empty())
+    }
+
+    #[uniffi::method(name = "setCustomBlockExplorer")]
+    pub fn _set_custom_block_explorer(&self, url: String) -> Result<()> {
+        if url.is_empty() {
+            return self.delete_custom_block_explorer();
+        }
+
+        self.set_custom_block_explorer(url)
+    }
+
+    #[uniffi::method(name = "clearCustomBlockExplorer")]
+    pub fn _clear_custom_block_explorer(&self) -> Result<()> {
+        self.delete_custom_block_explorer()
+    }
+
     pub(crate) fn get(&self, key: GlobalConfigKey) -> Result<Option<String>> {
         let read_txn =
             self.db.begin_read().map_err(|error| Error::DatabaseAccess(error.to_string()))?;
--- a/rust/src/database/global_config.rs
+++ b/rust/src/database/global_config.rs
@@ -60,9 +60,15 @@ impl From<GlobalConfigKey> for &'static str {
             GlobalConfigKey::DecoySelectedWalletId => "decoy_selected_wallet_id",
             GlobalConfigKey::LockedAt => "locked_at",
             GlobalConfigKey::OnboardingProgress => "onboarding_progress",
-            GlobalConfigKey::CustomBlockExplorer(Network::Bitcoin) => "custom_block_explorer_bitcoin",
-            GlobalConfigKey::CustomBlockExplorer(Network::Testnet) => "custom_block_explorer_testnet",
-            GlobalConfigKey::CustomBlockExplorer(Network::Testnet4) => "custom_block_explorer_testnet4",
+            GlobalConfigKey::CustomBlockExplorer(Network::Bitcoin) => {
+                "custom_block_explorer_bitcoin"
+            }
+            GlobalConfigKey::CustomBlockExplorer(Network::Testnet) => {
+                "custom_block_explorer_testnet"
+            }
+            GlobalConfigKey::CustomBlockExplorer(Network::Testnet4) => {
+                "custom_block_explorer_testnet4"
+            }
             GlobalConfigKey::CustomBlockExplorer(Network::Signet) => "custom_block_explorer_signet",
         }
     }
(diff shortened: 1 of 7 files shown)


rust/src/database/global_config.rs
```diff
@@ -302,6 +307,43 @@ impl GlobalConfigTable {
...
+        let url = self.get(key).unwrap_or(None).unwrap_or_default();
+        if url.is_empty() {
+            None
+        } else {
+            Some(url)
+        }
+    }
+
+    #[uniffi::method(name = "customBlockExplorer")]
+    pub fn _custom_block_explorer(&self, network: Network) -> Option<String> {
```
praveenperera: why `_` prefixed func names?

rust/src/transaction/transaction_details.rs
```diff
@@ -360,12 +370,14 @@ impl TransactionDetails {
...
-            Network::Testnet4 => format!("https://mempool.space/testnet4/tx/{}", self.tx_id.0),
-            Network::Signet => format!("https://mutinynet.com/tx/{}", self.tx_id.0),
+        if let Some(custom_url) = Database::global().global_config.custom_block_explorer(self.network) {
+            if !custom_url.is_empty() {
+                let base = custom_url.trim_end_matches('/');
+                return format!("{base}/{}", self.tx_id.0);
+            }
         }
+
+        Self::default_transaction_url(self.network, &self.tx_id)
```
praveenperera: what happens if they use a block explorer that doesn't use the same format `/tx/{}` for transactions?

praveenperera (changes requested):
w

praveenperera (changes requested):
Address earlier comments

rust/src/transaction/transaction_details.rs
Vineet1101: Initially I thought only of standard block explorers.now I added for non-standard ones

rust/src/database/global_config.rs
Vineet1101: there was a naming collision with other fn

rust/src/database/global_config.rs
```diff
@@ -302,6 +307,43 @@ impl GlobalConfigTable {
...
+                "Custom block explorer URL must start with http:// or https://".to_string(),
+            )));
+        }
+
+        let key = GlobalConfigKey::CustomBlockExplorer(network);
+        self.set(key, url)
+    }
+
+    #[uniffi::method(name = "clearCustomBlockExplorer")]
+    pub fn clear_custom_block_explorer(&self, network: Network) -> Result<()> {
```
praveenperera: `[uniffi::method(name` 

is not needed here the camelCasing happens automatically, only need this if the func name is not standard

rust/src/transaction/transaction_details.rs
```diff
@@ -360,12 +370,18 @@ impl TransactionDetails {
...
     pub fn transaction_url(&self) -> String {
-        match self.network {
-            Network::Bitcoin => format!("https://mempool.space/tx/{}", self.tx_id.0),
-            Network::Testnet => format!("https://mempool.space/testnet/tx/{}", self.tx_id.0),
-            Network::Testnet4 => format!("https://mempool.space/testnet4/tx/{}", self.tx_id.0),
-            Network::Signet => format!("https://mutinynet.com/tx/{}", self.tx_id.0),
+        if let Some(custom_url) = Database::global().global_config.custom_block_explorer(self.network) {
+            if !custom_url.is_empty() {
+                if custom_url.contains("{txid}") {
+                    return custom_url.replace("{txid}", &self.tx_id.0.to_string());
```
praveenperera: too much nesting

praveenperera (changes requested):

rust/src/database/global_config.rs
Vineet1101: okay

rust/src/transaction/transaction_details.rs
Vineet1101: fixed it

rust/src/transaction/transaction_details.rs
```diff
@@ -360,12 +370,22 @@ impl TransactionDetails {
...
+            return Self::default_transaction_url(self.network, &self.tx_id);
+        };
+
+        if url.is_empty() {
+            return Self::default_transaction_url(self.network, &self.tx_id);
+        }
+
+        if url.contains("{txid}") {
+            return url.replace("{txid}", &self.tx_id.0.to_string());
         }
```
praveenperera: I like this but i think we should also support `{tx_id}` and `{0}` (sparrow format)

praveenperera (changes requested):
This is good overall, just a small change, also did research on how sparrow handles this.


> In Sparrow, openBlockExplorer() checks whether the configured URL contains {0}. If it does, Sparrow replaces {0} with the txid. If it does not, it falls back to appending /<network>/tx/<txid> on test networks or /tx/<txid> on mainnet. See github.com/sparrowwallet/sparrow/src/main/java/com/sparrowwallet/sparrow/AppServices.java:941 and the settings text in github.com/sparrowwallet/sparrow/src/main/java/com/sparrowwallet/sparrow/settings/GeneralSettingsController.java:106.

We can merge this in like this, but we need to make it clear this is only the backend changes. More changes are needed for this to be a complete features. Including.

1. Backend functions to normalize different URLs (with/without scheme and path), if we get a URL without `{0}` `{txid}` etc, or without a path we should default to what sparrow does. We could also do a HTTP call to see if that pattern returns 200 for a known txn
2. We need to hook this up to the UI, add field somewhere, where the users can enter in a custom mempool/txn explorer link.

These can be done in future PRs

rust/src/transaction/transaction_details.rs
Vineet1101: done

Vineet1101:
> This is good overall, just a small change, also did research on how sparrow handles this.
> 
> > In Sparrow, openBlockExplorer() checks whether the configured URL contains {0}. If it does, Sparrow replaces {0} with the txid. If it does not, it falls back to appending //tx/ on test networks or /tx/ on mainnet. See github.com/sparrowwallet/sparrow/src/main/java/com/sparrowwallet/sparrow/AppServices.java:941 and the settings text in github.com/sparrowwallet/sparrow/src/main/java/com/sparrowwallet/sparrow/settings/GeneralSettingsController.java:106.
> 
> We can merge this in like this, but we need to make it clear this is only the backend changes. More changes are needed for this to be a complete features. Including.
> 
> 1. Backend functions to normalize different URLs (with/without scheme and path), if we get a URL without `{0}` `{txid}` etc, or without a path we should default to what sparrow does. We could also do a HTTP call to see if that pattern returns 200 for a known txn
> 2. We need to hook this up to the UI, add field somewhere, where the users can enter in a custom mempool/txn explorer link.
> 
> These can be done in future PRs

yup I agree. I am happy to implement this feature fully in future PRs

praveenperera (changes requested):
Regen bindings get CI passing

Vineet1101:
> Regen bindings get CI passing

I have regenerated the bindings. On doing so I realize that there is no specific cmd to regenerate just bindings part for both android and ios. As I am using linux machine I extracted the code from the just build-ios which is used to generate bindings and ran it manually. I think it is a good fix.

Do you want me to create a issue for this and then subsequently open a PR as this doesn't need too much of a work

praveenperera:
@Vineet1101 please get CI passing

praveenperera (changes requested):
The custom block explorer support has two correctness issues: common root-style explorer URLs generate broken transaction links, and backup restore bypasses the new URL validation entirely. Those cases make the feature unreliable and can persist non-web URLs that are later opened from the transaction details screens.


Insert /tx/ when no placeholder is configured `transaction/transaction_details.rs:405-406` When a user saves an explorer instance URL like https://mempool.space/testnet or a self-hosted mempool root, this fallback builds .../<txid> instead of .../tx/<txid>. The built-in defaults in default_transaction_url() all require the /tx/ segment, so View in Explorer will 404 for these accepted inputs unless the caller already knows to append /tx or provide a {txid} template.


Restore custom explorer URLs through the validated setter — backup/import.rs:600-601. If a backup contains custom_block_explorers, this writes them with config.set(...) instead of set_custom_block_explorer(...). That bypasses the new http(s) validation, so importing a crafted backup can persist schemes like sms: or intent: and later transaction_url() passes them straight to the transaction-details openURL/ACTION_VIEW flow. Using the setter here keeps restore behavior aligned with normal app input.

PR #650, opened 2026-04-16
add rescan with custom gap limit to recover missed addresses

guptapratykshh:
Adds a rescan action that takes a user-chosen gap limit, so wallets can find addresses with balances that were skipped past the default 30/150 limit. No ui yet, just the Rust hook (rescan_wallet_with_gap_limit) that the iOS and Android settings screens can wire up next.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous wallet rescan with a configurable gap limit that returns immediately and runs in background.
  * Rescans won't start if another full or incremental scan is already running.

* **Bug Fixes / Improvements**
  * Enforced gap-limit validation (must be within allowed range).
  * Wallet is marked as scanned when a rescan covers the expanded scan range.
  * Added debug logging for rescan requests.

### DIFF
--- a/rust/src/manager/wallet_manager.rs
+++ b/rust/src/manager/wallet_manager.rs
@@ -12,6 +12,7 @@ use parking_lot::RwLock;
 use tap::TapFallible as _;
 use tracing::{debug, error, warn};
 
+use cove_common::consts::MAX_RESCAN_GAP_LIMIT;
 use cove_tokio::task::{self, spawn_actor};
 use cove_util::{format::NumberFormatter as _, result_ext::ResultExt as _};
 
@@ -1001,6 +1002,24 @@ impl RustWalletManager {
         tokio::spawn(async move {
             send!(actor.wallet_scan_and_notify(true));
         });
+    }
+
+    #[uniffi::method]
+    pub async fn rescan_wallet_with_gap_limit(&self, gap_limit: u32) -> Result<(), Error> {
+        debug!("rescan_wallet_with_gap_limit: {} gap_limit={}", self.id, gap_limit);
+
+        if gap_limit == 0 || gap_limit > MAX_RESCAN_GAP_LIMIT {
+            return Err(Error::WalletScanError(format!(
+                "gap_limit must be between 1 and {MAX_RESCAN_GAP_LIMIT}",
+            )));
+        }
+
+        let actor = self.actor.clone();
+        task::spawn(async move {
+            send!(actor.perform_rescan_full_scan(gap_limit));
+        });
+
+        Ok(())
     }
 
     #[uniffi::method]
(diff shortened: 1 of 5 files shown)


rust/src/manager/wallet_manager.rs
```diff
@@ -1003,6 +1003,22 @@ impl RustWalletManager {
...
+    #[uniffi::method]
+    pub async fn rescan_wallet_with_gap_limit(&self, gap_limit: u32) -> Result<(), Error> {
+        debug!("rescan_wallet_with_gap_limit: {} gap_limit={}", self.id, gap_limit);
+
+        if gap_limit == 0 {
+            return Err(Error::WalletScanError("gap_limit must be >= 1".to_string()));
+        }
+
+        let actor = self.actor.clone();
+        tokio::spawn(async move {
```
praveenperera: use crate::task::spawn

praveenperera (changes requested):
This is a good start, but doesn't fix the issue fully since, we don't have a UI to call this. But this is a good start.

praveenperera:
also fix the commit message style read CONTRIBUTING.md

praveenperera:
@guptapratykshh thanks this is looking good can you also address the things found by coderabbit and greptile, if they don't apply reply to each and say why. thanks

rust/src/manager/wallet_manager.rs
```diff
@@ -1003,6 +1003,25 @@ impl RustWalletManager {
         });
     }
 
+    #[uniffi::method]
+    pub async fn rescan_wallet_with_gap_limit(&self, gap_limit: u32) -> Result<(), Error> {
+        debug!("rescan_wallet_with_gap_limit: {} gap_limit={}", self.id, gap_limit);
+
+        const MAX_RESCAN_GAP_LIMIT: u32 = 10_000;
```
praveenperera: lets put this either at the top of the file or with the other consts in the const file, also 10_000 seems like a lot

praveenperera (changes requested):

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -101,13 +101,23 @@ pub enum FullScanType {
...
             Self::Expanded => 150,
+            Self::Rescan(gap) => *gap as usize,
+        }
+    }
+
+    fn covers_expanded(&self) -> bool {
+        match self {
+            Self::Initial => false,
+            Self::Expanded => true,
+            Self::Rescan(_) => self.stop_gap() >= Self::Expanded.stop_gap(),
```
praveenperera: why not just use the gap in Rescan(gap) in here directly?, also add /// comments to these new functions

praveenperera (changes requested):

praveenperera (approved):

praveenperera:
Looks good thanks @guptapratykshh approved

praveenperera (changes requested):
need to regenerate bindings with `just bi` and `just ba` see the CONTRIBUTING.md file

Issue #657, opened 2026-04-16
Add canonical outpoint lock state and APIs


praveenperera:
Parent issue: #168

## Summary
Add the canonical per-outpoint lock model and the backend APIs needed by the app and UI layers.

This issue is about state and interfaces only. It should not include send-path enforcement logic, BIP329 sync behavior, or frontend UI work.

## Scope
- represent lock state by `OutPoint`
- add read/write APIs to lock and unlock a single outpoint
- add read APIs to inspect whether an outpoint is locked
- add bulk APIs or helper methods needed by the transaction-details shortcut to lock or unlock all wallet-owned unspent outputs for a transaction
- add any transaction-details aggregate state needed by the UI to render `unlocked | mixed | locked`
- scope transaction-level aggregation to **wallet-owned unspent outputs only**

## Out of scope
- excluding locked outpoints from transaction building
- BIP329 import/export sync
- Manage UTXOs UI
- Transaction Details UI

## Acceptance criteria
- backend lock state is represented canonically per outpoint
- callers can lock and unlock individual outpoints
- callers can query lock state for an outpoint
- callers can obtain aggregate transaction lock state over wallet-owned unspent outputs for a transaction
- callers can perform the transaction-details bulk action over wallet-owned unspent outputs for a transaction
- spent outputs are ignored by transaction-level aggregate and bulk operations

## Notes
This issue is intentionally the foundation for the other child issues. It should land first or provide the APIs they build on.

PR #664, opened 2026-04-16
Canonical UTXO lock state and backend APIs

Vineet1101:
### Platform Coverage

- [ ] Tested on iOS device
- [ ] Tested on iOS simulator
- [x] Not tested

## Checklist

- [✔️ ] I have read [CONTRIBUTING.md](https://github.com/bitcoinppl/cove/blob/master/CONTRIBUTING.md)
- [✔️ ] I have read [ARCHITECTURE.md](https://github.com/bitcoinppl/cove/blob/master/ARCHITECTURE.md)

I have relied on rust unit tests for this PR because this doesn't add any android or swift changes

## Summary by CodeRabbit

* **New Features**
  * Added a LockState (Unlocked, Mixed, Locked) exposed in APIs.
  * Lock and unlock individual or multiple transaction outputs; check per-output status and list locked outputs.
  * Transaction details now include a readable aggregate lock state and a method to toggle a transaction’s outputs between locked/unlocked.

* **Tests**
  * Unit tests covering lock/unlock behavior, listing, toggling, and mixed-state scenarios.
  
Fizes #657


rust/src/database/key/outpoint.rs
```diff
@@ -58,7 +58,7 @@ impl redb::Value for OutPointKey {
         Self: 'a,
     {
         let id = data[..32].try_into().expect("id is 32 bytes");
-        let index = u32::from_be_bytes(data[32..36].try_into().expect("index is 4 bytes"));
+        let index = u32::from_ne_bytes(data[32..36].try_into().expect("index is 4 bytes"));
```
Sandipmandal25, line 66: OutPointKey is already used by INPUT_TABLE and OUTPUT_TABLE in label.rs. Changing from_be_bytes to from_ne_bytes silently changes deserialization for those existing tables too. Safe right now only because label.rs discards keys and never reads .index back, but that's a fragile assumption. This should be its own commit with a comment explaining why the existing tables are unaffected and what breaks if someone starts reading key indices from them.

praveenperera (changes requested):
I don’t think this PR is quite landing on the right boundary yet.

The new `locked_outpoints` storage itself seems reasonable, but the higher-level integration does not.

What feels off is the API surface:

- `rust/src/utxo_lock_manager.rs` is basically a thin wrapper over the new table. It doesn’t really behave like the other UI-facing managers in the codebase, and it pushes tx-level assembly work onto the caller.
- If lock state is supposed to be shown in Transaction Details, then the backend should expose it as part of `TransactionDetails`, not require the UI to fetch transaction details, separately figure out which outpoints matter, and then ask another object for aggregate state.
- Right now `TransactionDetails` still has no `lock_state`, and `WalletActor::transaction_details(...)` still doesn’t compute or attach one.

I also don’t think this fully satisfies issue #657 yet:

- The issue requires transaction-level aggregate state for Transaction Details.
- The issue requires tx-level aggregate/bulk operations to be scoped to wallet-owned unspent outputs only.
- The current code leaves that filtering up to the caller by accepting arbitrary `Vec<OutPoint>`.
- That means spent outputs are not actually ignored by the backend; they’re only ignored if the caller does the right thing.

Suggested direction:

- Keep `rust/src/database/wallet_data/locked_outpoints.rs` as a dedicated table under `wallet_data`.
- Remove or at least avoid introducing a standalone `UtxoLockManager` as the main integration point.
- Add `lock_state: LockState` to `TransactionDetails`.
- Compute that state inside the existing transaction-details path using wallet-owned unspent outputs for the transaction.
- Expose tx-centric operations like `toggle_transaction_lock_state(tx_id)` from the wallet/actor layer, where the backend can enforce the “wallet-owned unspent outputs only” rule itself.
- Add backend tests proving transaction-level aggregate/toggle ignores spent outputs and only considers wallet-owned unspent outputs


So in short: new table in `wallet_data` makes sense, new database does not, separate top-level manager probably does not, and the missing piece is integrating this into the existing Transaction Details backend path so it actually addresses `#657`.

rust/src/manager/wallet_manager.rs
```diff
@@ -346,6 +346,16 @@ impl RustWalletManager {
...
 
+    #[uniffi::method]
+    pub fn utxo_lock_manager(
+        &self,
+    ) -> Result<Arc<crate::utxo_lock_manager::UtxoLockManager>, Error> {
+        let manager = crate::utxo_lock_manager::UtxoLockManager::new(self.id.clone())
+            .map_err(|e| Error::UnknownError(e.to_string()))?;
+
+        Ok(Arc::new(manager))
+    }
```
Sandipmandal25, line 369: i think each call to this creates a new WalletDataDb and opens a write transaction to initialize tables, which could block other writers. also UtxoLockManager has a `#[uniffi::constructor]` that lets you construct it without going through wallet_manager, so there are two entry points now  not sure if that's intentional? wondering if caching the manager in RustWalletManager and returning the same Arc would be cleaner

rust/src/utxo_lock_manager.rs
```diff
@@ -0,0 +1,257 @@
...
+                self.lock_all(outpoints)?;
+                Ok(LockState::Locked)
+            }
+            LockState::Locked => {
+                self.unlock_all(outpoints)?;
+                Ok(LockState::Unlocked)
+            }
+        }
+    }
+}
```
Sandipmandal25: `aggregate_lock_state` and `lock_all/unlock_all` are two separate DB transactions so state can change between the read and the write. thinking if the user taps toggle twice fast the second call might read stale state, would wrapping both in a single write transaction make this atomic?

Sandipmandal25 (changes requested):
@praveenperera took a look at this left a few questions inline. main things i wasn't sure about the `from_ne_bytes` change touches OutPointKey which is shared with label.rs, `toggle_lock_state` doing two separate transactions, and` utxo_lock_manager()` potentially being called multiple times. curious what you think!

devvaansh:
() already implements redb::Value natively, so wrapping it in Cbor<()> adds serialization overhead for a zero-information value. This is a presence/absence table , we only care whether the key exists. Using TableDefinition<OutPointKey, ()> directly would be simpler and more efficient. So for future you should consider this!

rust/src/database/key/outpoint.rs
praveenperera: yes why was this changed @vineet1101?

Vineet1101:
> I don’t think this PR is quite landing on the right boundary yet.
> 
> The new `locked_outpoints` storage itself seems reasonable, but the higher-level integration does not.
> 
> What feels off is the API surface:
> 
> * `rust/src/utxo_lock_manager.rs` is basically a thin wrapper over the new table. It doesn’t really behave like the other UI-facing managers in the codebase, and it pushes tx-level assembly work onto the caller.
> * If lock state is supposed to be shown in Transaction Details, then the backend should expose it as part of `TransactionDetails`, not require the UI to fetch transaction details, separately figure out which outpoints matter, and then ask another object for aggregate state.
> * Right now `TransactionDetails` still has no `lock_state`, and `WalletActor::transaction_details(...)` still doesn’t compute or attach one.
> 
> I also don’t think this fully satisfies issue #657 yet:
> 
> * The issue requires transaction-level aggregate state for Transaction Details.
> * The issue requires tx-level aggregate/bulk operations to be scoped to wallet-owned unspent outputs only.
> * The current code leaves that filtering up to the caller by accepting arbitrary `Vec<OutPoint>`.
> * That means spent outputs are not actually ignored by the backend; they’re only ignored if the caller does the right thing.
> 
> Suggested direction:
> 
> * Keep `rust/src/database/wallet_data/locked_outpoints.rs` as a dedicated table under `wallet_data`.
> * Remove or at least avoid introducing a standalone `UtxoLockManager` as the main integration point.
> * Add `lock_state: LockState` to `TransactionDetails`.
> * Compute that state inside the existing transaction-details path using wallet-owned unspent outputs for the transaction.
> * Expose tx-centric operations like `toggle_transaction_lock_state(tx_id)` from the wallet/actor layer, where the backend can enforce the “wallet-owned unspent outputs only” rule itself.
> * Add backend tests proving transaction-level aggregate/toggle ignores spent outputs and only considers wallet-owned unspent outputs
> 
> So in short: new table in `wallet_data` makes sense, new database does not, separate top-level manager probably does not, and the missing piece is integrating this into the existing Transaction Details backend path so it actually addresses `#657`.

yup I agree with you I have addressed your feedback in the new code

praveenperera:
@Vineet1101 there is merge conflicts, please resolve

rust/src/database/wallet_data/locked_outpoints.rs
```diff
@@ -0,0 +1,384 @@
...
+
+        {
+            let mut table = write_txn.open_table(LOCKED_OUTPOINTS_TABLE)?;
+            let key = OutPointKey::from(outpoint);
+            table.remove(key)?;
+        }
+
+        write_txn.commit().map_err_str(DatabaseError::DatabaseAccess)?;
+        Ok(())
+    }
```
Sandipmandal25, line 67: locked entries are never removed when the output gets spent. behavior is correct since the actor filters to unspent, but stale rows accumulate with no cleanup path worth noting this as a known gap for whoever wires up coin control later.

rust/src/manager/cloud_backup_manager.rs
```diff
@@ -1791,6 +1791,7 @@ impl RustCloudBackupManager {
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::await_holding_lock)]
```
Sandipmandal25, line 1984: adding a module level allow here silences this lint for all future tests too, isnt this unreleted to this pr?

rust/src/manager/cloud_backup_manager/ops.rs
```diff
@@ -1098,6 +1098,7 @@ mod test_support;
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::await_holding_lock)]
```
Sandipmandal25, line 1185: same thing as earlier i think these need to fixed in their own pr.

Sandipmandal25 (commented):
added two small things.

rust/src/manager/cloud_backup_manager.rs
praveenperera: its fine, we if we don't add these CI won't pass, i'm okay with it for tests, if its a problem the test will fail

praveenperera:
seems abandoned, closing, you can open a new one with merge conflicts fixed if you wish

PR #665, opened 2026-04-16
Add wallet recovery word status tracking and error handling

### LINKED ISSUE #316: Need better "missing seed words" warning
I ran into a weird situation when my Cove wallet had to be restored on to a new phone.  The Apple tech broke the original phone while replacing a failed battery. After restoring apps to the new phone Cove still showed the balance for my hot wallet using the public key descriptors with no indication the original seed was now missing. A few months later when I went to spend from that wallet or view the seed words I get the below error (in a black font on dark background): 
```
CoveCore.MnemonicError.NotAvailable(
"3IoXviHNOpaL|soS84uAV")
```

I understand now that the mnemonic is stored in the iOS Secure Enclave which can't be restored to a new device. But it would be nice if Cove showed an indication on the "My Wallets" list when a hot wallet's seed words are missing so the user will be prompted sooner to restore the wallet before sending any more BTC to a potentially lost key. 

Image

kartikangiras:
## Summary
Display a clear warning in My Wallets when a hot wallet's recovery words are unavailable on device. This addresses the issue where users restore to a new iPhone and don't realize their seed words are missing from the Secure Enclave which doesn't transfer between devices.

Fixes #316 
## Testing
- just build-ios
- just ci
- just fmt

### Platform Coverage

- [ ] Tested on iOS device
- [x] Tested on iOS simulator
- [ ] Not tested
Recovery words warning only appears when keychain entry is missing which happens in device restore. Could not simulate this in simulator without manual keychain deletion.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error messaging for recovery words with improved visual styling, including color-coded alerts for better clarity.
  * Fixed wallet item layout in sidebar to prevent text overflow and added border styling.
  * QR code button now displays only when recovery words are available.


ios/Cove/Flows/SelectedWalletFlow/SecretWordsScreen.swift
```diff
@@ -26,6 +27,21 @@ struct SecretWordsScreen: View {
         (words?.words().count ?? 24) / numberOfColumns
     }
 
+    var recoveryWordsErrorMessage: String {
+        if let loadError {
```
praveenperera: combine if let with switch just match directly in switch

ios/Cove/AppManager.swift
```diff
@@ -13,6 +13,7 @@ private let walletModeChangeDelayMs = 250
     var router: Router
     var database: Database
     var wallets: [WalletMetadata] = []
+    var walletsMissingRecoveryWords: Set<WalletId> = []
```
praveenperera: why do we need this and refreshing recovery words status on each change to wallets in AppManager? Just check if theres an error in the SecretWordsScreen

praveenperera (changes requested):

ios/Cove/Flows/SelectedWalletFlow/SecretWordsScreen.swift
```diff
@@ -15,7 +15,8 @@ struct SecretWordsScreen: View {
 
     // private
     @State var words: Mnemonic?
-    @State var errorMessage: String?
+    @State private var loadError: MnemonicError?
+    @State private var fallbackErrorMessage: String?
```
praveenperera: can you explain why we have 2 error messages here? can we use an enum? and we need to  bring the change to android too.

praveenperera (changes requested):

ios/Cove/Flows/SelectedWalletFlow/SecretWordsScreen.swift
kartikangiras: my initial thought was to have one err check for rust and other the normal try catch, but it was a redundant approach so i shifted to using enum and also applied similar changes to android as well.

praveenperera:
@kartikangiras can you respond to my comments inline, and resolve or respond to the bots too thanks

ios/Cove/Flows/SelectedWalletFlow/SecretWordsScreen.swift
kartikangiras: done

ios/Cove/Flows/SelectedWalletFlow/SecretWordsScreen.swift
```diff
@@ -58,7 +74,18 @@ struct SecretWordsScreen: View {
...
+                        .foregroundStyle(.white.opacity(0.95))
+                        .padding(16)
+                        .frame(maxWidth: screenWidth * 0.9, alignment: .leading)
+                        .background(Color.red.opacity(0.16))
+                        .overlay {
+                            RoundedRectangle(cornerRadius: 10)
+                                .stroke(Color.red.opacity(0.30), lineWidth: 1)
+                        }
+                        .cornerRadius(10)
                 }
```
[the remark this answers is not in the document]
kartikangiras: resolved

ios/Cove/Flows/SelectedWalletFlow/SecretWordsScreen.swift
```diff
@@ -26,6 +27,19 @@ struct SecretWordsScreen: View {
...
+        switch loadError {
+        case .some(.NotAvailable):
+            "Recovery words are unavailable on this device for this wallet. This can happen after restoring to a new iPhone because Secure Enclave data does not transfer. Restore this wallet from your saved recovery words before sending more BTC to it."
+        case .some(.GetWalletKeychain):
+            "Cove could not access this wallet's recovery words in the keychain right now."
+        case .some(.UnknownWord):
+            "Cove could not decode this wallet's recovery words."
+        case .none:
+            fallbackErrorMessage ?? "Loading..."
+        }
```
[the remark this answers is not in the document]
kartikangiras: done

ios/Cove/AppManager.swift
kartikangiras: yeah indeed, i just simplified to do a regular error check in SecretWordsScreen

kartikangiras:
@praveenperera I have address all the inline review comments and from bots as well.

android/app/src/main/java/org/bitcoinppl/cove/secret_words/SecretWordsScreen.kt
```diff
@@ -208,9 +215,15 @@ fun SecretWordsScreen(
                         contentAlignment = Alignment.Center,
                     ) {
                         Text(
-                            text = errorMessage ?: stringResource(R.string.label_loading),
-                            color = Color.White.copy(alpha = 0.7f),
+                            text = loadError?.message ?: stringResource(R.string.label_loading),
```
praveenperera, line 218: `loadError?.message ?: stringResource(R.string.label_loading)` is rendered inside the same red error-styled `Text`, so the normal initial loading state still appears as an error on Android while `Mnemonic(id = walletId)` is running. Split this branch like the iOS version: show a neutral loading view when `loadError == null`, and only apply the red background/padding to actual `RecoveryWordsLoadError` messages.

praveenperera (changes requested):

praveenperera:
Closing this for now after maintainer triage. The branch is conflicting, CI is not green, and the PR scope/title no longer matches the current diff. Please reopen or send a fresh PR after rebasing, fixing the Android loading/error state, and aligning the scope with the implementation.

PR #666, opened 2026-04-16
Feat cache receive addresses with `5m timeout`

### LINKED ISSUE #386: Show the same address in a 5 min window
Right now every time you click the receive address or hit Copy on the address we show a new address, but this makes it hard for people to confirm the address they pasted in is the same address as before. This was  done to reduce the amount of accidental address re-use, but this causes a different issue.

Fix:

1. Add a timeout of 5 mins, only show a new address if its been 5 mins since an address was shown
2. Always show a new address if theres a transaction pending or confirmed on the previous address (pretty sure BDK handles this already but confirm)

oyetanishq:
## Summary
1. Previously, the receive address refreshed every time the view was opened or the address was copied.
2. Closes: #386  

This commit introduces address caching to the WalletManager on both platforms:
1. Addresses are now cached for `5 minutes` before a new one is requested.
2. The cache is automatically invalidated if the wallet's `transaction count changes`, ensuring a genuinely new address is shown if the current one receives pending or confirmed funds.
3. Users can still `manually force a new address` generation via the "Create New Address" button.

## Testing
|IOS|Android|
|-|-|
|Screenshot 2026-04-16 at 21 37 18|WhatsApp Image 2026-04-17 at 00 01 52|

### Platform Coverage

- [x] Tested on iOS device
- [x] Tested on Android device
- [x] Tested on iOS simulator
- [x] Tested on Android simulator
- [ ] Not tested

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Receive addresses auto-refresh and expire after 5 minutes on Android and iOS.
  * Live countdown in the UI: shows "Auto-refresh in XX" during the final minute, then "Refreshing..." and requests a new address.
  * "Create new address" forces an immediate refresh.
  * Intelligent caching limits unnecessary regenerations while ensuring timely updates.

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/flows/SelectedWalletFlow/ReceiveAddressSheet.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/flows/SelectedWalletFlow/ReceiveAddressSheet.kt
@@ -67,18 +67,20 @@ fun ReceiveAddressSheet(
     var addressInfo by remember { mutableStateOf<AddressInfoWithDerivation?>(null) }
     var isLoading by remember { mutableStateOf(true) }
     var errorMessage by remember { mutableStateOf<String?>(null) }
+    var initialLoadComplete by remember { mutableStateOf(false) }
 
-    // load initial address on mount
-    LaunchedEffect(manager) {
+    // load initial address on mount, re-fetch safely on state changes
+    LaunchedEffect(manager, manager.loadState) {
         try {
             isLoading = true
-            addressInfo = manager.rust.nextAddress()
+            addressInfo = manager.getReceiveAddress(forceNew = false)
             errorMessage = null
         } catch (e: Exception) {
             Log.e(tag, "Unable to get next address", e)
             errorMessage = e.message ?: "Unable to get address"
         } finally {
             isLoading = false
+            initialLoadComplete = true
         }
     }
 
@@ -87,7 +89,7 @@ fun ReceiveAddressSheet(
         scope.launch {
             try {
                 isLoading = true
-                addressInfo = manager.rust.nextAddress()
+                addressInfo = manager.getReceiveAddress(forceNew = true)
                 errorMessage = null
             } catch (e: Exception) {
                 Log.e(tag, "Unable to get next address", e)
@@ -133,8 +135,10 @@ fun ReceiveAddressSheet(
             addressRaw = addressInfo?.addressUnformatted(),
             derivationPath = addressInfo?.derivationPath(),
             isLoading = isLoading,
+            addressGeneratedTime = manager.addressGeneratedTime,
             onCopyAddress = ::copyAddress,
             onCreateNewAddress = ::createNewAddress,
+            initialLoadComplete = initialLoadComplete,
         )
     }
 }
@@ -146,10 +150,39 @@ private fun ReceiveAddressSheetContent(
     addressRaw: String?,
     derivationPath: String?,
     isLoading: Boolean,
+    addressGeneratedTime: Long,
+    initialLoadComplete: Boolean,
     onCopyAddress: () -> Unit,
     onCreateNewAddress: () -> Unit,
 ) {
     val isDarkTheme = !MaterialTheme.colorScheme.isLight
+    var timeLeftStr by remember { mutableStateOf("") }
+
+    LaunchedEffect(addressGeneratedTime, initialLoadComplete) {
+        if (!initialLoadComplete) return@LaunchedEffect
+        while(true) {
+            if (addressGeneratedTime == 0L) {
+                timeLeftStr = ""
+            } else {
+                val elapsed = System.currentTimeMillis() - addressGeneratedTime
+                val remaining = (5 * 60 * 1000L) - elapsed
+                
+                if (remaining > 0) {
+                    if (remaining <= 60_000L) {
+                        val seconds = (remaining / 1000)
+                        timeLeftStr = String.format("Auto-refresh in %02d", seconds)
+                    } else {
+                        timeLeftStr = ""
+                    }
+                } else {
+                    timeLeftStr = "Refreshing..."
+                    onCreateNewAddress()
+                    break
+                }
+            }
+            kotlinx.coroutines.delay(1000)
+        }
+    }
 
     Column(
         modifier =
@@ -278,7 +311,18 @@ private fun ReceiveAddressSheetContent(
             }
         }
 
-        Spacer(modifier = Modifier.height(64.dp))
+        Spacer(modifier = Modifier.height(32.dp))
+
+        if (timeLeftStr.isNotEmpty() && !isLoading) {
+            Text(
+                text = timeLeftStr,
+                style = MaterialTheme.typography.bodySmall,
+                color = MaterialTheme.colorScheme.onSurfaceVariant.copy(alpha = 0.6f),
+                modifier = Modifier.padding(top = 4.dp)
+            )
+        }
+
+        Spacer(modifier = Modifier.height(32.dp))
 
         // copy address button
         Button(
@@ -330,6 +374,8 @@ private fun ReceiveAddressSheetPreview() {
             isLoading = false,
             onCopyAddress = {},
             onCreateNewAddress = {},
+            addressGeneratedTime = System.currentTimeMillis(),
+            initialLoadComplete = false,
         )
     }
 }
--- a/ios/Cove/Flows/SelectedWalletFlow/ReceiveView.swift
+++ b/ios/Cove/Flows/SelectedWalletFlow/ReceiveView.swift
@@ -20,6 +20,39 @@ struct ReceiveView: View {
     private let pasteboard = UIPasteboard.general
     @State private var addressInfo: AddressInfoWithDerivation?
 
+    @State private var timeRemainingString: String = ""
+    @State private var timerTask: Task<Void, Never>? = nil
+
+    func startTimer() {
+        timerTask?.cancel()
+        timerTask = Task { @MainActor in
+            while !Task.isCancelled {
+                guard let generatedTime = manager.addressGeneratedTime else {
+                    timeRemainingString = ""
+                    try? await Task.sleep(for: .seconds(1))
+                    continue
+                }
+
+                let elapsed = Date().timeIntervalSince(generatedTime)
+                let remaining = (5 * 60) - elapsed
+
+                if remaining > 0 {
+                    if remaining <= 60 {
+                        let seconds = Int(remaining)
+                        timeRemainingString = String(format: "Auto-refresh in %02d", seconds)
+                    } else {
+                        timeRemainingString = ""
+                    }
+                } else {
+                    timeRemainingString = "Refreshing..."
+                    nextAddressSync()
+                    break
+                }
+                try? await Task.sleep(for: .seconds(1))
+            }
+        }
+    }
+
     var addressLoaded: Bool {
         addressInfo != nil
     }
@@ -38,13 +71,16 @@ struct ReceiveView: View {
     }
 
     func nextAddressSync() {
-        Task { await nextAddress() }
+        Task { await nextAddress(forceNew: true) }
     }
 
-    func nextAddress() async {
+    func nextAddress(forceNew: Bool = false) async {
         do {
-            let addressInfo = try await manager.rust.nextAddress()
-            await MainActor.run { self.addressInfo = addressInfo }
+            let addressInfo = try await manager.getReceiveAddress(forceNew: forceNew)
+            await MainActor.run {
+                self.addressInfo = addressInfo
+                self.startTimer()
+            }
         } catch {
             Log.error("Unable to get next address: \(error)")
             dismiss()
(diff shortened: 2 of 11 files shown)


praveenperera (changes requested):
good start and thanks for the screenshots, We are only right now saving this in memory, so if the user exits and comes back, this will be lost. And a new address will be generated.So let's use the redb cache table. to save this in on on disk. So handle the save as much of the logic as possible in Rust. Read architecture.md. Also only show the countdown timer if it is within one minute of expiring and make that text uh read w but at some opacity.

ios/Cove/Flows/SelectedWalletFlow/ReceiveView.swift
```diff
@@ -20,6 +20,38 @@ struct ReceiveView: View {
...
+                    if remaining <= 60 {
+                        let seconds = Int(remaining)
+                        timeRemainingString = String(format: "Auto-refresh in %02d", seconds)
+                    } else {
+                        timeRemainingString = ""
+                    }
+                } else {
+                    timeRemainingString = "Address Expired"
+                    break
+                }
```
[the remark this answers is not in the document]
oyetanishq: those image are of older commit, in current version only last minute is shown

praveenperera (changes requested):
Looking good @oyetanishq, 2 things
   
  1. **wallet_manager.rs:903-905**: The new cache invalidation logic treats any transaction-count change as if the displayed receive address was used. Because both receive screens now call getReceiveAddress(false) whenever loadState changes, unrelated wallet activity such as initial sync progress or an outgoing payment can rotate the receive address even when the cached address is still unused. That causes the QR code/address to change unexpectedly and can burn through the unused-address window.  

  2.  **ReceiveView.swift:46-48**: When the five-minute countdown reaches zero, the timer only sets "Address Expired" and exits. It never requests a fresh receive address. If the receive screen stays open past expiry and no loadState update happens, the user continues seeing and potentially copying a stale address even though the UI says it will auto-refresh. The Android countdown appears to have the same issue.

Also we want to make sure that IF the current address does get used we auto switch to the next address even if the timer is not up

oyetanishq:
### Updates:
1. Now when BDK's next address's index is greater than current cached address's index, it request address update.
2. After 5 minutes, address updates instead of "Address Expired"

oyetanishq:
heyy @praveenperera can you please review this pr!!

rust/src/manager/wallet_manager.rs
```diff
@@ -879,6 +885,62 @@ impl RustWalletManager {
...
+
+            let mut address_was_used = false;
+            if current_tx_count != cache.tx_count {
+                let next_address =
+                    call!(self.actor.next_address()).await.map_err_str(Error::NextAddressError)?;
+
+                // If BDK's next unused address index is greater than cached index
+                // It means cached address received funds and BDK moved the pointer
+                address_was_used = next_address.index() > cache.address_info.index();
+            }
```
praveenperera, line 914: This validation path advances wallet state. When `current_tx_count != cache.tx_count`, calling `self.actor.next_address()` here is not a read-only check: `wallet.get_next_address()` reveals and persists a fresh external address whenever the gap window is not full. 

That means an unrelated transaction-count change, like creating a send, can burn an unseen receive address even though this method still returns the cached address. This should use a non-mutating check instead of advancing wallet state during cache validation.

Look at `peek_address`

rust/src/manager/wallet_manager.rs
```diff
@@ -879,6 +885,62 @@ impl RustWalletManager {
...
+                }
+
+                self.reconciler.send(Message::AddressGeneratedTime(cache.generated_at));
+                return Ok(cache.address_info);
+            }
+        }
+
+        let new_address =
+            call!(self.actor.next_address()).await.map_err_str(Error::NextAddressError)?;
+        let now = jiff::Timestamp::now().as_second() as u64;
```
praveenperera, line 932: This can skip the immediate next receive address after the cached one is used. If `current_tx_count != cache.tx_count`, the earlier `self.actor.next_address()` call has already advanced BDK to the next valid receive address. Calling `next_address()` again here advances a second time, so the UI ends up showing the address after that and the immediate successor is never presented. This can happen as soon as a receive to the cached address increases the wallet’s transaction count before the sheet reloads.

praveenperera (changes requested):

rust/src/manager/wallet_manager.rs
oyetanishq: Well looking at the function `peek_address`, I thought of another way to store address in cache, then check if the stored address is still available in unused addresses list, if yes then no need to update, otherwise get `next_address`.

rust/src/manager/wallet_manager.rs
oyetanishq: should we remove this `tx_count`??, ig its adding unnecessary complexity, and just check if address is present in unused addresses?

android/app/src/main/java/org/bitcoinppl/cove/flows/SelectedWalletFlow/ReceiveAddressSheet.kt
```diff
@@ -146,10 +147,37 @@ private fun ReceiveAddressSheetContent(
...
+                if (remaining > 0) {
+                    if (remaining <= 60_000L) {
+                        val seconds = (remaining / 1000)
+                        timeLeftStr = String.format("Auto-refresh in %02d", seconds)
+                    } else {
+                        timeLeftStr = ""
+                    }
+                } else {
+                    timeLeftStr = "Refreshing..."
+                    onCreateNewAddress()
```
praveenperera, line 179: This timer can call `onCreateNewAddress()` while the initial `LaunchedEffect(manager, manager.loadState)` is still fetching the receive address. If `addressGeneratedTime` is stale from a previous sheet, the initial fetch may already refresh the expired cache, then this forced refresh reveals another address immediately afterward. That skips the first generated address and consumes gap-limit space on reopen. Gate the timer until the current address has loaded, or have the initial fetch handle expiration and start the countdown only after the refreshed timestamp arrives.

ios/Cove/Flows/SelectedWalletFlow/ReceiveView.swift
```diff
@@ -20,6 +20,39 @@ struct ReceiveView: View {
...
+                if remaining > 0 {
+                    if remaining <= 60 {
+                        let seconds = Int(remaining)
+                        timeRemainingString = String(format: "Auto-refresh in %02d", seconds)
+                    } else {
+                        timeRemainingString = ""
+                    }
+                } else {
+                    timeRemainingString = "Refreshing..."
+                    nextAddressSync()
```
praveenperera, line 48: This timer can fire before the initial `.task { await nextAddress() }` has completed when `manager.addressGeneratedTime` is stale from a previous receive sheet. The initial load already refreshes an expired cache via `getReceiveAddress(forceNew: false)`; if this forced call queues behind it, the app reveals a second address immediately, skipping the first new receive address and unnecessarily consuming gap-limit space. Gate auto-refresh until an address has been loaded for the current sheet/timestamp, or let the initial load own expiry and only start the timer after the new generated time is published.

praveenperera (changes requested):
I found one cross-platform issue: reopening receive after the cached timestamp is stale can trigger both the initial address load and the timer refresh, revealing two addresses back-to-back.

ios/Cove/Flows/SelectedWalletFlow/ReceiveView.swift
```diff
@@ -38,13 +71,16 @@ struct ReceiveView: View {
...
 
-    func nextAddress() async {
+    func nextAddress(forceNew: Bool = false) async {
         do {
-            let addressInfo = try await manager.rust.nextAddress()
-            await MainActor.run { self.addressInfo = addressInfo }
+            let addressInfo = try await manager.getReceiveAddress(forceNew: forceNew)
+            await MainActor.run {
+                self.addressInfo = addressInfo
+                self.startTimer()
```
praveenperera, line 82: `startTimer()` reads `manager.addressGeneratedTime`, which is delivered asynchronously through reconcile and may still contain the previous sheet's expired timestamp when this new address load completes. On reopen, the initial `getReceiveAddress(forceNew: false)` can already generate the fresh address, then this timer immediately sees the stale expired timestamp and calls `nextAddressSync()`, forcing a second address and skipping the one that was just displayed. Start the timer only after the timestamp for the current load has been applied, or return the generated timestamp with the address and drive the countdown from that paired value.

android/app/src/main/java/org/bitcoinppl/cove/flows/SelectedWalletFlow/ReceiveAddressSheet.kt
```diff
@@ -146,10 +150,39 @@ private fun ReceiveAddressSheetContent(
...
     isLoading: Boolean,
+    addressGeneratedTime: Long,
+    initialLoadComplete: Boolean,
     onCopyAddress: () -> Unit,
     onCreateNewAddress: () -> Unit,
 ) {
     val isDarkTheme = !MaterialTheme.colorScheme.isLight
+    var timeLeftStr by remember { mutableStateOf("") }
+
+    LaunchedEffect(addressGeneratedTime, initialLoadComplete) {
```
praveenperera, line 161: `initialLoadComplete` flips after `getReceiveAddress(forceNew = false)` returns, but `manager.addressGeneratedTime` is still updated through an async reconcile path and can still be the previous expired sheet timestamp. That lets this effect immediately call `onCreateNewAddress()` after the initial fetch already refreshed the cache, which reveals a second address and skips the first new receive address. Gate the countdown on a timestamp that belongs to the loaded address, or have `getReceiveAddress` return both the address and generated time so the timer never runs against stale manager state.

rust/src/database/global_cache.rs
```diff
@@ -60,6 +71,43 @@ pub enum GlobalCacheTableError {
 }
 
 impl GlobalCacheTable {
+    pub fn get_receive_address(&self, id: &WalletId) -> Result<Option<ReceiveAddressCache>, Error> {
+        let read_txn = self.db.begin_read().map_err_str(Error::DatabaseAccess)?;
+        let table = read_txn.open_table(TABLE).map_err_str(Error::TableAccess)?;
+
+        let id_str: &str = id.as_ref();
+        let key = format!("receive_address_{}", id_str);
```
praveenperera, line 79: The receive-address cache key only includes `wallet_id`, so it survives address-type switches and wallet deletion. After switching descriptors, `get_receive_address` can return a cached address from the previous address type for up to five minutes; after deletion, the global cache row is orphaned. Include the active address type in the key or explicitly delete this cache entry when switching address types, and add deletion cleanup in the wallet-delete flow.

rust/src/manager/wallet_manager.rs
```diff
@@ -880,6 +886,63 @@ impl RustWalletManager {
...
+    /// Get the receive address for the wallet, deduplicated and persistently cached
+    #[uniffi::method]
+    pub async fn get_receive_address(
+        &self,
+        force_new: bool,
+    ) -> Result<AddressInfoWithDerivation, Error> {
+        let _guard = self.address_generation_lock.lock().await;
+        let db = Database::global();
+
+        let txns = call!(self.actor.transactions()).await.unwrap_or_default();
```
praveenperera, line 898: Every receive-address request builds and sorts the full transaction list just to call `.len()`, and this happens while holding the address-generation lock. On larger wallets or repeated `loadState` changes, opening the receive sheet can do unnecessary transaction work before returning an address. Add a lightweight actor method that counts the same filtered transactions without allocating `Transaction` objects or sorting.

rust/src/manager/wallet_manager.rs
```diff
@@ -880,6 +886,63 @@ impl RustWalletManager {
...
+            call!(self.actor.next_address()).await.map_err_str(Error::NextAddressError)?;
+        let now = jiff::Timestamp::now().as_second() as u64;
+
+        let cache = crate::database::global_cache::ReceiveAddressCache {
+            address_info: new_address.clone(),
+            generated_at: now,
+            tx_count: current_tx_count,
+        };
+
+        let _ = db.global_cache.set_receive_address(&self.id, cache);
```
praveenperera, line 940: This cache write, and the similar tx-count update write above, silently discard database errors. If persistence fails, the app still returns an address but nothing is cached, so the next receive-sheet load can rotate again with no diagnostic trail. Log the error with the existing `tap_err` pattern or propagate it if the cache write is required for the reuse guarantee.

android/app/src/main/java/org/bitcoinppl/cove/flows/SelectedWalletFlow/ReceiveAddressSheet.kt
```diff
@@ -146,10 +150,39 @@ private fun ReceiveAddressSheetContent(
...
+            if (addressGeneratedTime == 0L) {
+                timeLeftStr = ""
+            } else {
+                val elapsed = System.currentTimeMillis() - addressGeneratedTime
+                val remaining = (5 * 60 * 1000L) - elapsed
+                
+                if (remaining > 0) {
+                    if (remaining <= 60_000L) {
+                        val seconds = (remaining / 1000)
+                        timeLeftStr = String.format("Auto-refresh in %02d", seconds)
```
praveenperera, line 173: This new `String.format` uses the device default locale, unlike the surrounding Android codebase which passes `Locale.US` for numeric formatting. That can trigger the existing implicit-locale lint and make the countdown formatting locale-dependent. Import `java.util.Locale` and call `String.format(Locale.US, "Auto-refresh in %02d", seconds)`.

praveenperera (changes requested):
The non-mutating unused-address validation addresses the earlier `next_address()` cache-validation concern, but the receive-sheet timer can still force-refresh against a stale timestamp on both platforms.

praveenperera:
@oyetanishq also fix merge conflicts

PR #670, opened 2026-04-16
 fix: skip unrecognized BIP329 label entries instead of failing the w…

### LINKED ISSUE #277: [BUG] Error Importing Labels from Nunchuck
Image

Sandipmandal25:
## Summary

Fixes #277

When importing BIP329 labels from Nunchuck, a single entry with an unrecognized ref format (e.g. non-numeric `vout`) caused the entire import to fail with `"Unable to parse file: error parsing vout"`, resulting in all labels being lost.

`parse_labels` now processes input line-by-line: valid entries are imported, invalid ones are logged and skipped. The import only fails if every line is unrecognized.

## Testing

5 unit tests added covering:

* all valid
* mixed valid/invalid
* all invalid
* empty input
* blank lines

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Label import now returns a parsed report (imported vs skipped) and UIs surface partial-import details.
  * Backup import reports now include skipped label counts.

* **Bug Fixes**
  * Import parsing tolerates blank lines and skips malformed/unsupported entries instead of failing the whole import.
  * Transaction refresh after imports remains intact.

* **Tests**
  * Added tests for partial success, full failure, empty input, and blank-line handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bitcoinppl/cove/pull/670)

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/nfc/NfcLabelImportSheet.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/nfc/NfcLabelImportSheet.kt
@@ -45,6 +45,7 @@ fun NfcLabelImportSheet(
     labelManager: LabelManager,
     onDismiss: () -> Unit,
     onSuccess: () -> Unit,
+    onPartialSuccess: (skipped: UInt) -> Unit,
     onError: (String) -> Unit,
 ) {
     val context = LocalContext.current
@@ -86,11 +87,13 @@ fun NfcLabelImportSheet(
                     // try to parse the NFC data as BIP329 labels
                     try {
                         result.text?.let { text ->
-                            // try to import the labels
-                            labelManager.import(text.trim())
-                            // success! dismiss and notify
+                            val importResult = labelManager.import(text.trim())
                             nfcReader.reset()
-                            onSuccess()
+                            if (importResult.skipped > 0u) {
+                                onPartialSuccess(importResult.skipped)
+                            } else {
+                                onSuccess()
+                            }
                             return@collect
                         }
 
--- a/android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
@@ -298,14 +298,15 @@ internal fun WalletSheetsHost(
                 onPartialSuccess = { skipped ->
                     onDismissNfcScanner()
                     scope.launch {
-                        val noun = if (skipped == 1u) "entry" else "entries"
-                        val msg = "Labels imported ($skipped $noun skipped)"
+                        val noun = if (skipped == 1u) "label" else "labels"
+                        val skippedMsg = "Labels imported. Could not import $skipped $noun"
                         try {
                             manager.rust.getTransactions()
+                            snackbarHostState.showSnackbar(skippedMsg)
                         } catch (e: Exception) {
                             android.util.Log.e(tag, "Failed to refresh transactions after partial NFC label import")
+                            snackbarHostState.showSnackbar("$skippedMsg. Transaction list may need manual refresh")
                         }
-                        snackbarHostState.showSnackbar(msg)
                     }
                 },
                 onError = { errorMsg ->
--- a/android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
@@ -295,6 +295,19 @@ internal fun WalletSheetsHost(
                         }
                     }
                 },
+                onPartialSuccess = { skipped ->
+                    onDismissNfcScanner()
+                    scope.launch {
+                        val noun = if (skipped == 1u) "entry" else "entries"
+                        val msg = "Labels imported ($skipped $noun skipped)"
+                        try {
+                            manager.rust.getTransactions()
+                        } catch (e: Exception) {
+                            android.util.Log.e(tag, "Failed to refresh transactions after partial NFC label import")
+                        }
+                        snackbarHostState.showSnackbar(msg)
+                    }
+                },
                 onError = { errorMsg ->
                     onDismissNfcScanner()
                     scope.launch {
--- a/android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
@@ -304,7 +304,7 @@ internal fun WalletSheetsHost(
                             manager.rust.getTransactions()
                             snackbarHostState.showSnackbar(skippedMsg)
                         } catch (e: Exception) {
-                            android.util.Log.e(tag, "Failed to refresh transactions after partial NFC label import")
+                            android.util.Log.e(tag, "Failed to refresh transactions after partial NFC label import", e)
                             snackbarHostState.showSnackbar("$skippedMsg. Transaction list may need manual refresh")
                         }
                     }
--- a/ios/Cove/Flows/SelectedWalletFlow/SelectedWalletScreen.swift
+++ b/ios/Cove/Flows/SelectedWalletFlow/SelectedWalletScreen.swift
@@ -448,13 +448,24 @@ struct SelectedWalletScreen: View {
         }
 
         do {
-            try labelManager.importLabels(labels: labels)
-            app.alertState = .init(
-                .general(
-                    title: "Success!",
-                    message: "Labels have been imported successfully."
+            let result = try labelManager.importLabels(labels: labels)
+            if result.skipped > 0 {
+                app.alertState = .init(
+                    .general(
+                        title: "Labels Imported",
+                        message: "Could not import \(result.skipped) labels"
+                    )
                 )
-            )
+            } else {
+                app.alertState = .init(
+                    .general(
+                        title: "Success!",
+                        message: "Labels have been imported successfully."
+                    )
+                )
+            }
+
+            Task { await manager.rust.getTransactions() }
 
         } catch {
             app.alertState = .init(
--- a/ios/Cove/Flows/SelectedWalletFlow/SelectedWalletScreen.swift
+++ b/ios/Cove/Flows/SelectedWalletFlow/SelectedWalletScreen.swift
@@ -368,14 +368,24 @@ struct SelectedWalletScreen: View {
             do {
                 let file = try result.get()
                 let fileContents = try FileReader(for: file).read()
-                try labelManager.import(jsonl: fileContents)
+                let result = try labelManager.import(jsonl: fileContents)
 
-                app.alertState = .init(
-                    .general(
-                        title: "Success!",
-                        message: "Labels have been imported successfully."
+                if result.skipped > 0 {
+                    let noun = result.skipped == 1 ? "label" : "labels"
+                    app.alertState = .init(
+                        .general(
+                            title: "Labels Imported",
+                            message: "Could not import \(result.skipped) \(noun)"
+                        )
                     )
-                )
+                } else {
+                    app.alertState = .init(
+                        .general(
+                            title: "Success!",
+                            message: "Labels have been imported successfully."
+                        )
+                    )
+                }
 
                 // when labels are imported, we need to get the transactions again with the updated labels
                 Task { await manager.rust.getTransactions() }
@@ -450,10 +460,11 @@ struct SelectedWalletScreen: View {
         do {
             let result = try labelManager.importLabels(labels: labels)
             if result.skipped > 0 {
+                let noun = result.skipped == 1 ? "label" : "labels"
                 app.alertState = .init(
                     .general(
                         title: "Labels Imported",
-                        message: "Could not import \(result.skipped) labels"
+                        message: "Could not import \(result.skipped) \(noun)"
                     )
                 )
             } else {
--- a/rust/src/backup/import.rs
+++ b/rust/src/backup/import.rs
@@ -534,7 +534,14 @@ pub(crate) fn cleanup_failed_wallet(metadata: &WalletMetadata) -> Vec<String> {
 
 fn import_labels(id: &WalletId, jsonl: &str) -> Result<(), BackupError> {
     let manager = LabelManager::new(id.clone());
-    manager.import(jsonl).map(|_| ()).map_err_str(BackupError::Restore)
+    manager.import(jsonl).map_err_str(BackupError::Restore)?;
+    Ok(())
+}
+
+fn import_labels_preserve_backup(id: &WalletId, jsonl: &str) -> Result<(), BackupError> {
+    let manager = LabelManager::new(id.clone());
+    manager.import_without_cloud_backup_dirty(jsonl).map_err_str(BackupError::Restore)?;
+    Ok(())
 }
 
 pub(crate) fn restore_wallet_labels(
@@ -547,13 +554,11 @@ pub(crate) fn restore_wallet_labels(
         return LabelRestoreOutcome::default();
     };
 
-    let manager = LabelManager::new(wallet_id.clone());
     let import_result = match behavior {
         LabelRestoreBehavior::MarkCloudBackupDirty => import_labels(wallet_id, jsonl),
-        LabelRestoreBehavior::PreserveCloudBackupClean => manager
-            .import_without_cloud_backup_dirty(jsonl)
-            .map(|_| ())
-            .map_err_str(BackupError::Restore),
+        LabelRestoreBehavior::PreserveCloudBackupClean => {
+            import_labels_preserve_backup(wallet_id, jsonl)
+        }
     };
 
     match import_result {
--- a/rust/src/backup/import.rs
+++ b/rust/src/backup/import.rs
@@ -533,7 +533,7 @@ pub(crate) fn cleanup_failed_wallet(metadata: &WalletMetadata) -> Vec<String> {
 
 fn import_labels(id: &WalletId, jsonl: &str) -> Result<(), BackupError> {
     let manager = LabelManager::new(id.clone());
-    manager.import(jsonl).map_err(|e| BackupError::Restore(e.to_string()))
+    manager.import(jsonl).map(drop).map_err(|e| BackupError::Restore(e.to_string()))
 }
 
 pub(crate) fn restore_wallet_labels(
@@ -551,6 +551,7 @@ pub(crate) fn restore_wallet_labels(
         LabelRestoreBehavior::MarkCloudBackupDirty => import_labels(wallet_id, jsonl),
         LabelRestoreBehavior::PreserveCloudBackupClean => manager
             .import_without_cloud_backup_dirty(jsonl)
+            .map(drop)
             .map_err(|error| BackupError::Restore(error.to_string())),
     };
 
--- a/rust/src/backup/import.rs
+++ b/rust/src/backup/import.rs
@@ -11,7 +11,7 @@ use cove_types::network::Network;
 
 use crate::database::Database;
 use crate::database::global_config::GlobalConfigKey;
-use crate::label_manager::LabelManager;
+use crate::label_manager::{LabelManager, LabelParseReport};
 use crate::mnemonic::MnemonicExt as _;
 use crate::wallet::fingerprint::Fingerprint;
 use crate::wallet::metadata::{WalletId, WalletMetadata, WalletMode, WalletType};
@@ -43,6 +43,7 @@ pub async fn import_all(
             Ok(RestoreResult::Imported {
                 name,
                 labels_imported,
+                labels_skipped,
                 labels_failure,
                 fingerprint,
                 degraded,
@@ -51,6 +52,7 @@ pub async fn import_all(
                 if labels_imported {
                     report.wallets_with_labels_imported += 1;
                 }
+                report.labels_skipped += labels_skipped;
                 if let Some((name, error)) = labels_failure {
                     report.labels_failed_wallet_names.push(name);
                     report.labels_failed_errors.push(error);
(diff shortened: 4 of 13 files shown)


rust/src/label_manager.rs
```diff
@@ -532,5 +533,89 @@ impl LabelManager {
...
+    let mut skipped = 0usize;
+
+    for line in &lines {
+        match serde_json::from_str::<Label>(line) {
+            Ok(label) => parsed.push(label),
+            Err(e) => {
+                if first_error.is_none() {
+                    first_error = Some(e.to_string());
+                }
+                warn!("skipping unrecognized label entry: {e}");
```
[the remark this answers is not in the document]
Sandipmandal25: done!

rust/src/label_manager.rs
```diff
@@ -532,5 +533,89 @@ impl LabelManager {
...
+        let jsonl = format!("{VALID_TX}\n{VALID_OUTPUT}\n{VALID_INPUT}");
+        let labels = parse_labels(&jsonl).unwrap();
+        assert_eq!(labels.len(), 3);
+    }
+
+    #[test]
+    fn test_parse_labels_skips_bad_vout_line() {
+        // simulates the Nunchuck import bug: one line with an unparseable vout
+        // the good lines must still be imported
+        let jsonl = format!("{VALID_TX}\n{BAD_VOUT_OUTPUT}\n{VALID_OUTPUT}");
```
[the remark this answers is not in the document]
Sandipmandal25: fixed!

rust/src/label_manager.rs
```diff
@@ -532,5 +533,89 @@ impl LabelManager {
...
+                warn!("skipping unrecognized label entry: {e}");
+                skipped += 1;
+            }
+        }
+    }
+
+    if parsed.is_empty() {
+        return Err(LabelManagerError::Parse(
+            first_error.unwrap_or_else(|| "no valid labels found".to_string()),
+        ));
```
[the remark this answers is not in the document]
Sandipmandal25: fixed!

praveenperera:
@Sandipmandal25 this looks good thanks, 2 questions before approving

1. do we know why nunchuk uses this format? is it valid? should we just make ours less strict?
2. should we report back to the user how many rows could not be imported? maybe even give reasons why?

Sandipmandal25:
> @Sandipmandal25 this looks good thanks, 2 questions before approving
> 
> 1. do we know why nunchuk uses this format? is it valid? should we just make ours less strict?

 we dont have a real nunchuk export to verify the exact format the issue mentioned coin collection labels with custom hashtag types but didnt shared a sanitized file. what we do know is bip329 spec explicitly says "importing wallets complying to this specification should ignore types not defined here", so making Cove less strict isnt a workaround its what the spec requires. the old `Labels::try_from_str` was the non compliant behavior. 

> 2. should we report back to the user how many rows could not be imported? maybe even give reasons why?

agreed right now skips only go to the debug log. surfacing a count + reasons to the user means changing import to return something richer than Result which touches the Uniffi  boundary and needs ui changes to show the warning. can  to add it here or track it as a follow up issue whichever you prefer.

praveenperera:
@Sandipmandal25 lets do the full change including sending all the info back, but for now lets keep the UI simple and just say `could not import xx labels` but lets make a new issue so we don't forget to hook up the new report to the UI to show the reasons

Sandipmandal25:
> @Sandipmandal25 lets do the full change including sending all the info back, but for now lets keep the UI simple and just say `could not import xx labels` but lets make a new issue so we don't forget to hook up the new report to the UI to show the reasons

sounds good updating the pr shortly! thanks

rust/src/label_manager.rs
```diff
@@ -532,16 +548,16 @@ impl LabelManager {
     }
 }
 
-fn parse_labels(jsonl: &str) -> Result<Labels, LabelManagerError> {
+fn parse_labels(jsonl: &str) -> Result<(Labels, u32), LabelManagerError> {
```
praveenperera: shouldnt we just return LabelImportResult here too?

lets think of a better name and use it everywhere mabye LabelParseReport ? Then later we can add reasons to this struct

praveenperera (changes requested):

rust/src/label_manager.rs
Sandipmandal25: missed it renamed to labelparsereport and using it in parse labels too. also regenerated ios and android bindings

rust/src/backup/import.rs
```diff
@@ -524,7 +524,7 @@ pub(crate) fn cleanup_failed_wallet(metadata: &WalletMetadata) -> Vec<String> {
 
 fn import_labels(id: &WalletId, jsonl: &str) -> Result<(), BackupError> {
     let manager = LabelManager::new(id.clone());
-    manager.import(jsonl).map_err(|e| BackupError::Restore(e.to_string()))
+    manager.import(jsonl).map(drop).map_err(|e| BackupError::Restore(e.to_string()))
```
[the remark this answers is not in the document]
Sandipmandal25: agreed but the restore path already has its own warning system. will handle in a follow up if needed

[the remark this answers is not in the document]
Sandipmandal25: no thanks, will open one manually if needed

rust/src/label_manager.rs
```diff
@@ -313,21 +327,24 @@ impl LabelManager {
...
+    ) -> Result<LabelImportResult, LabelManagerError> {
+        let imported = labels.len() as u32;
         self.db.labels.insert_labels(labels).map_err_str(LabelManagerError::Save)?;
         if mark_cloud_backup_dirty {
             self.mark_cloud_backup_dirty();
         }
 
-        Ok(())
+        Ok(LabelImportResult { imported, skipped })
     }
```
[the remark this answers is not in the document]
Sandipmandal25: only skipped is shown in the UI right now so this is fine will revisit if imported gets surfaced

android/app/src/main/java/org/bitcoinppl/cove/nfc/NfcLabelImportSheet.kt
```diff
@@ -86,11 +86,13 @@ fun NfcLabelImportSheet(
...
-                            labelManager.import(text.trim())
-                            // success! dismiss and notify
+                            val importResult = labelManager.import(text.trim())
                             nfcReader.reset()
-                            onSuccess()
+                            if (importResult.skipped > 0u) {
+                                onError("Could not import ${importResult.skipped} labels")
+                            } else {
+                                onSuccess()
+                            }
```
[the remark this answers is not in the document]
Sandipmandal25: fixed plural. keeping onerror for now per @praveenperera  guidance can refine in a follow up

praveenperera (changes requested):
I found three correctness issues that still need to be addressed before this is ready:

- `rust/src/label_manager.rs:318-320`  
  `importLabels()` reports all parsed labels as imported because it uses `labels.len()`, but `LabelsTable::insert_label_with_write_txn()` still drops unsupported BIP329 variants like `pubkey` and `xpub` in its `_` arm. That means scanner imports can show full success even when some records were never persisted. The import report should be computed after filtering to the label types the wallet can actually store.

- `android/app/src/main/java/org/bitcoinppl/cove/AppManager.kt:355-356`  
  The skipped-label handling still does not apply to QR/camera imports with a malformed JSONL line. `MultiFormat::try_from_string()` builds `Bip329Labels` with the strict `bip329::Labels::try_from_str` parser, so those scans fail before `importLabels(multiFormat.v1)` is ever reached. Scanner imports should also go through the tolerant parser so valid labels are still imported.

- `android/app/src/main/java/org/bitcoinppl/cove/nfc/NfcLabelImportSheet.kt:91-94`  
  Partial NFC imports are still reported as failures. This code calls `onSuccess()` and then immediately calls `onError()` when some labels are skipped. `WalletSheetsHost` treats `onError` as a hard failure, so the user gets a `Failed to import labels...` snackbar even though labels were imported and transactions were refreshed. Partial imports need a success-with-warning path instead of reusing the failure callback.

Sandipmandal25:
> I found three correctness issues that still need to be addressed before this is ready:
> 
> * `rust/src/label_manager.rs:318-320`
>   `importLabels()` reports all parsed labels as imported because it uses `labels.len()`, but `LabelsTable::insert_label_with_write_txn()` still drops unsupported BIP329 variants like `pubkey` and `xpub` in its `_` arm. That means scanner imports can show full success even when some records were never persisted. The import report should be computed after filtering to the label types the wallet can actually store.
> * `android/app/src/main/java/org/bitcoinppl/cove/AppManager.kt:355-356`
>   The skipped-label handling still does not apply to QR/camera imports with a malformed JSONL line. `MultiFormat::try_from_string()` builds `Bip329Labels` with the strict `bip329::Labels::try_from_str` parser, so those scans fail before `importLabels(multiFormat.v1)` is ever reached. Scanner imports should also go through the tolerant parser so valid labels are still imported.
> * `android/app/src/main/java/org/bitcoinppl/cove/nfc/NfcLabelImportSheet.kt:91-94`
>   Partial NFC imports are still reported as failures. This code calls `onSuccess()` and then immediately calls `onError()` when some labels are skipped. `WalletSheetsHost` treats `onError` as a hard failure, so the user gets a `Failed to import labels...` snackbar even though labels were imported and transactions were refreshed. Partial imports need a success-with-warning path instead of reusing the failure callback.

Hey @praveenperera  addressed all LabelParseReport with imported/skipped counts is wired up across Rust iOS and Android. All import paths (QR, NFC, file) now go through the tolerant line by line parser and surface partial import warnings to the user. Also hit a pretty large merge conflict during rebase since this overlapped with #685  took a bit to ort out cleanly. Should be good for another look when you get a chance!

Sandipmandal25:
also squashed the commits into 1 to keep the repo clean.

praveenperera:
thanks but don't squash anymore @Sandipmandal25, makes it harder to  review just the changes made since last review, thanks

rust/src/backup/import.rs
praveenperera: why the `drop` here?

rust/src/backup/import.rs
```diff
@@ -542,6 +542,7 @@ pub(crate) fn restore_wallet_labels(
         LabelRestoreBehavior::MarkCloudBackupDirty => import_labels(wallet_id, jsonl),
         LabelRestoreBehavior::PreserveCloudBackupClean => manager
             .import_without_cloud_backup_dirty(jsonl)
+            .map(drop)
```
praveenperera: and here?

rust/src/label_manager.rs
```diff
@@ -231,13 +238,17 @@ impl LabelManager {
...
 
     #[uniffi::method(name = "importLabels")]
-    pub fn _import_labels(&self, labels: Arc<Bip329Labels>) -> Result<(), LabelManagerError> {
+    pub fn _import_labels(
```
praveenperera: lets just call this `ffi_import_labels`

rust/src/label_manager.rs
```diff
@@ -299,28 +310,47 @@ impl LabelManager {
...
+    ) -> Result<LabelParseReport, LabelManagerError> {
+        let labels = labels.into();
+        // Count only supported variants — insert_label_with_write_txn silently drops
+        // PublicKey and ExtendedPublicKey, so labels.len() would overstate the import count.
+        let supported = labels
+            .iter()
+            .filter(|l| {
+                matches!(
+                    l,
+                    Label::Transaction(_) | Label::Address(_) | Label::Input(_) | Label::Output(_)
```
praveenperera: lets extract this matches into a variable the communicates meaning

rust/src/label_manager.rs
```diff
@@ -299,28 +310,47 @@ impl LabelManager {
...
 
-    pub fn import_labels(&self, labels: impl Into<Labels>) -> Result<(), LabelManagerError> {
-        self.save_imported_labels(labels.into(), true)
+    pub fn import_labels(
+        &self,
+        labels: impl Into<Labels>,
+    ) -> Result<LabelParseReport, LabelManagerError> {
+        let labels = labels.into();
+        // Count only supported variants — insert_label_with_write_txn silently drops
+        // PublicKey and ExtendedPublicKey, so labels.len() would overstate the import count.
```
praveenperera: nit: the codebase convention is to start inline comments (non doc comments) with lowercase

praveenperera (changes requested):
looks, good almost ready, just some small changes and some questions.

rust/src/backup/import.rs
Sandipmandal25: import returns a report but this function just needs success or failure so i used map drop to throw away the report. changed it now to make it clearer.

rust/src/backup/import.rs
Sandipmandal25: same here

rust/src/label_manager.rs
Sandipmandal25: thanks done

rust/src/label_manager.rs
Sandipmandal25: lowercased them

rust/src/label_manager.rs
Sandipmandal25: done!

Sandipmandal25:
> looks, good almost ready, just some small changes and some questions.

done the chnages maybe you can took a final look.

rust/src/multi_format.rs
```diff
@@ -142,9 +142,15 @@ impl MultiFormat {
...
-            return Ok(Self::Bip329Labels(Arc::new(labels.into())));
+        // try and parse bip329 labels — tolerant: skip unparseable lines so a single
+        // malformed entry doesn't discard all valid labels in the file
+        let parsed_labels: Vec<bip329::Label> = string
+            .trim()
+            .lines()
+            .filter_map(|line| bip329::Label::try_from_str(line.trim()).ok())
+            .collect();
+        if !parsed_labels.is_empty() {
+            return Ok(Self::Bip329Labels(Arc::new(bip329::Labels::new(parsed_labels).into())));
```
praveenperera: These lines drop malformed JSONL rows before the data reaches `LabelManager`, so every import path that goes through `MultiFormat::Bip329Labels` loses the skipped-row count. With a payload that contains one valid label and one malformed line, the import now partially succeeds but `result.skipped` stays `0`, so the scan/file-open flows still show full success even though one entry was ignored. Please preserve the raw JSONL for import, or carry a parse report/skipped count alongside `Bip329Labels`, and add a regression test for a mixed valid+invalid scan payload.

ios/Cove/Flows/SelectedWalletFlow/SelectedWalletScreen.swift
```diff
@@ -429,13 +429,24 @@ struct SelectedWalletScreen: View {
...
+                        message: "Could not import \(result.skipped) labels"
+                    )
                 )
-            )
+            } else {
+                app.alertState = .init(
+                    .general(
+                        title: "Success!",
+                        message: "Labels have been imported successfully."
+                    )
```
praveenperera, line 475: This QR path now handles `LabelParseReport`, but the `.fileImporter` block earlier in this same view still discards the return value from `labelManager.import(jsonl:)` and always shows the full-success alert. That means the standard iOS “Import Labels” flow still claims everything imported even when some rows were skipped. Mirror the `result.skipped` handling there so file imports behave the same as the QR/NFC paths.

rust/src/label_manager.rs
```diff
@@ -299,28 +313,45 @@ impl LabelManager {
...
+        let is_supported = |l: &&Label| {
+            matches!(
+                l,
+                Label::Transaction(_) | Label::Address(_) | Label::Input(_) | Label::Output(_)
+            )
+        };
+        let supported = labels.iter().filter(is_supported).count() as u32;
+        let skipped = labels.len() as u32 - supported;
+        let report = LabelParseReport { imported: supported, skipped };
+        self.save_imported_labels(labels, report, true)
```
praveenperera, line 338: `import_labels()` now returns success even when every parsed label is unsupported (`imported == 0`, `skipped > 0`). In that case we still call `save_imported_labels`, nothing is actually persisted, the UI says “Labels Imported”, and we mark the cloud backup dirty despite importing nothing. This should mirror `parse_labels()` and return a parse error when no supported labels remain after filtering.

praveenperera (changes requested):
I found three actionable issues in the diff. The main one is that `MultiFormat` still drops malformed BIP329 rows before `LabelManager` sees them, so the scan/file-open paths that import `MultiFormat::Bip329Labels` can partially import labels while still reporting full success. There is also an iOS file-import path that still ignores the new `LabelParseReport`, and `import_labels()` currently treats unsupported-only payloads as a successful import. I could not verify prior PR discussion in this environment because `prc` was unavailable and both requested `gh` calls failed to reach GitHub, so this review is based on the checked-out diff.

rust/src/multi_format.rs
Sandipmandal25: done also regression test added for mixed valid+invalid scan payload, used the same txid already in the label manager test fixtures.

ios/Cove/Flows/SelectedWalletFlow/SelectedWalletScreen.swift
Sandipmandal25: done. mirrored the same skipped handling in the fileImporter block

rust/src/label_manager.rs
Sandipmandal25: added early return with a parse error when supported count is zero, so nothing gets persisted and cloud backup stays clean

rust/src/backup/import.rs
```diff
@@ -524,7 +525,7 @@ pub(crate) fn cleanup_failed_wallet(metadata: &WalletMetadata) -> Vec<String> {
 
 fn import_labels(id: &WalletId, jsonl: &str) -> Result<(), BackupError> {
     let manager = LabelManager::new(id.clone());
-    manager.import(jsonl).map_err(|e| BackupError::Restore(e.to_string()))
+    manager.import(jsonl).map(|_| ()).map_err_str(BackupError::Restore)
```
praveenperera: lol wtf is this, ugly as hell

rust/src/backup/import.rs
```diff
@@ -542,7 +543,8 @@ pub(crate) fn restore_wallet_labels(
         LabelRestoreBehavior::MarkCloudBackupDirty => import_labels(wallet_id, jsonl),
         LabelRestoreBehavior::PreserveCloudBackupClean => manager
             .import_without_cloud_backup_dirty(jsonl)
-            .map_err(|error| BackupError::Restore(error.to_string())),
+            .map(|_| ())
```
praveenperera: wtf?

praveenperera (changes requested):

rust/src/backup/import.rs
Sandipmandal25: fixed the ugly chaining.

rust/src/backup/import.rs
Sandipmandal25: fixed!

android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
```diff
@@ -295,6 +295,20 @@ internal fun WalletSheetsHost(
...
+                onPartialSuccess = { skipped ->
+                    onDismissNfcScanner()
+                    scope.launch {
+                        val noun = if (skipped == 1u) "label" else "labels"
+                        val skippedMsg = "Labels imported. Could not import $skipped $noun"
+                        try {
+                            manager.rust.getTransactions()
+                            snackbarHostState.showSnackbar(skippedMsg)
+                        } catch (e: Exception) {
+                            android.util.Log.e(tag, "Failed to refresh transactions after partial NFC label import")
```
praveenperera: This catch path still drops the exception object, so a failed transaction refresh after a partial NFC label import only leaves a generic log line and no stack trace or cause. Pass the caught `e` into `Log.e`, matching the nearby import paths, so the refresh failure is debuggable:

```kotlin
android.util.Log.e(tag, "Failed to refresh transactions after partial NFC label import", e)
```

praveenperera (changes requested):

android/app/src/main/java/org/bitcoinppl/cove/wallet/WalletSheets.kt
Sandipmandal25: resloved!

rust/src/backup/import.rs
```diff
@@ -524,7 +525,14 @@ pub(crate) fn cleanup_failed_wallet(metadata: &WalletMetadata) -> Vec<String> {
 
 fn import_labels(id: &WalletId, jsonl: &str) -> Result<(), BackupError> {
     let manager = LabelManager::new(id.clone());
-    manager.import(jsonl).map_err(|e| BackupError::Restore(e.to_string()))
+    manager.import(jsonl).map_err_str(BackupError::Restore)?;
```
praveenperera: `manager.import(jsonl)` now returns a report with `skipped`, but this restore path discards it and later returns `LabelRestoreOutcome { imported: true, warning: None }`. With the tolerant parser, a backup containing valid labels plus malformed or unsupported rows restores partially while the backup report says everything was clean. Preserve the report here and in the preserve-backup path, then thread `skipped > 0` into `LabelRestoreOutcome`/`LabelRestoreWarning` or `BackupImportReport` so backup restores surface partial label imports.

praveenperera (changes requested):
backup restore still discards the new LabelParseReport, so partial label restores can look fully clean while rows were skipped.

rust/src/backup/import.rs
Sandipmandal25: fixed this both import paths now preserve the parse report instead of discarding it. partial label restores will now surface the skipped count all the way up to the backup import report so a restore that looks clean can no longer silently hide skipped rows. and solving the conflicts now

PR #675, opened 2026-04-17
Fix ShareLink in remaining transient presenters

pradhyum6144:
Apply the same fix from #631 to the two remaining ShareLink sites that present from transient contexts.

**MoreInfoPopover (high risk):** `DownloadBackupButton` used a `ShareLink` inside a toolbar `Menu`. Menus are transient presenters — the same dialog-dismissal race from #449/#313 applies. Replaced with a `Button` that calls `ShareSheet.present` deferred by 400ms, matching the pattern from #631.

**TapSignerSetupSuccessView (lower risk):** The Download Backup button also used `ShareLink`. This screen is not a dialog, so the race is less likely, but replacing with `ShareSheet.present` eliminates the fragile path and keeps the sharing behaviour consistent across the app.

`SendFlowHardwareScreen:444` was already fixed in #631.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced backup sharing functionality to improve reliability and stability when accessing share options from menu interfaces
  * Improved error handling for share operations to better report any issues during backup export

### DIFF
--- a/ios/Cove/Flows/SelectedWalletFlow/MoreInfoPopover.swift
+++ b/ios/Cove/Flows/SelectedWalletFlow/MoreInfoPopover.swift
@@ -46,11 +46,7 @@ struct MoreInfoPopover: View {
         Task {
             do {
                 let result = try await manager.rust.exportTransactionsCsv()
-                ShareSheet.present(data: result.content, filename: result.filename) { success in
-                    if !success {
-                        Log.warn("Transaction Export Failed: cancelled or failed")
-                    }
-                }
+                ShareSheet.presentFromMenu(data: result.content, filename: result.filename)
             } catch {
                 app.alertState = .init(.general(
                     title: "Transaction Export Failed",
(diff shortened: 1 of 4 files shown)


praveenperera (approved):
Great thank tyou @pradhyum6144 !

pradhyum6144:
> Great thank tyou @pradhyum6144 !
Hi @praveenperera I’m interested in contributing to this project. If there are any open issues suitable for me I’d be happy to work on one. Thank you!

praveenperera:
@pradhyum6144 I think this is worth doing actually, centralize it and the delay

> In `@ios/Cove/Flows/SelectedWalletFlow/MoreInfoPopover.swift`:
- Around line 82-91: This duplicates a 0.4s magic delay in the Button action
around ShareSheet.present; extract and centralize the defer logic into a single
helper (e.g. ShareSheet.presentFromMenu(data:filename:) or a shared constant
MenuDismissalDelay used by a new helper) and update this Button (and the
exportTransactions caller) to call that helper so the delay, comment/rationale
and the existing failure-logging behavior (ShareSheet.present { success in ...
}) are kept in one place and not repeated.

praveenperera (approved):
approved thanks!

PR #688, opened 2026-04-19
Add wallet reordering to sidebar (iOS + Android)

### LINKED ISSUE #493: Add ability to sort wallets in hamburger menu
Love Cove — an extremely promising project! Well done!

A usability feature that would be cool is the ability to sort wallets in the hamburger menu that displays the list of wallets.

Envisaged user journey: tap hamburger icon to see wallets > tap and hold on wallet user wants to move, and drag up and down to sort into desired order.

diyaayay:
## Summary

Adds wallet reordering in the hamburger menu (closes #493).

- **Rust**: `position: u32` on `WalletMetadata`, stable-sorted `get_all()`, one-time `migrate_positions()` backfill on startup.
- **UniFFI**: `reorderWallets(orderedIds)` and `moveWallet(walletId, toPosition)`; new `WalletTableError::{WalletNotFound, ReorderMismatch}`.
- **iOS**: drag-to-reorder in `SidebarView` using `List` + `.onMove` + `EditMode`; haptics on drag start and on successful persist; optimistic update with rollback on FFI error; edit-mode resets when the sidebar closes.
- **Android**: long-press drag-to-reorder in `SidebarView.kt`; haptics; visual elevation (scale + shadow + zIndex); smooth `animateItem()` for neighbors; FFI persisted off the UI thread; optimistic update with rollback on error.
- **Bindings**: regenerated Kotlin and Swift UniFFI bindings.

## Testing

- `cargo fmt --check` — pass
- `cargo clippy --all-features -- -D warnings` — pass
- `cargo test --package cove --lib` — passes aside from pre-existing `database::migration::bdk` / `database::encrypted_backend` flakiness (reproduced on `master`; unrelated to this PR)
- `just ba` — Android build + Kotlin bindings regenerated successfully
- `just bi` — iOS build + Swift bindings regenerated successfully

Manual: reordered wallets on an Android device (Pixel) and on the iOS simulator; killed and reopened the app — wallet order persisted in the new position on both.

### Platform Coverage

- [ ] Tested on iOS device
- [x] Tested on Android device
- [x] Tested on iOS simulator
- [ ] Not tested

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Android: long-press drag-and-drop to reorder wallets in the sidebar.
  * iOS: “Edit Order” mode with drag-reorder, Done control, and contextual edit action.

* **Improvements**
  * Ordering is now persisted and consistently applied across devices; background migration repairs positions.
  * Smoother visuals, elevation/translation effects and enhanced haptic feedback during reordering.

* **Bug Fixes**
  * Optimistic reorders now rollback or reconcile if persistence fails, with reload on mismatch.

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
@@ -85,8 +85,7 @@ fun SidebarView(
     val scope = rememberCoroutineScope()
     val spacingPx = with(density) { INTER_ITEM_SPACING_DP.dp.toPx() }
 
-    // local, optimistically-updated order of wallets for a smooth drag; synced from
-    // `app.wallets` whenever no drag is in progress
+    // local copy lets drags apply optimistically; resyncs from app.wallets when idle
     var walletList by remember { mutableStateOf(app.wallets) }
     var draggedWalletId by remember { mutableStateOf<WalletId?>(null) }
     var draggedOffsetY by remember { mutableFloatStateOf(0f) }
@@ -156,7 +155,7 @@ fun SidebarView(
 
         Spacer(modifier = Modifier.height(12.dp))
 
-        // wallet list with long-press drag-to-reorder
+        // wallet list
         LazyColumn(
             modifier = Modifier.weight(1f),
             verticalArrangement = Arrangement.spacedBy(INTER_ITEM_SPACING_DP.dp),
--- a/android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
@@ -56,6 +56,7 @@ import androidx.compose.ui.unit.sp
 import androidx.compose.ui.zIndex
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import org.bitcoinppl.cove.AppManager
@@ -90,6 +91,8 @@ fun SidebarView(
     var draggedWalletId by remember { mutableStateOf<WalletId?>(null) }
     var draggedOffsetY by remember { mutableFloatStateOf(0f) }
     var itemHeightPx by remember { mutableFloatStateOf(0f) }
+    var preDragSnapshot by remember { mutableStateOf<List<WalletMetadata>>(emptyList()) }
+    var pendingReorderJob by remember { mutableStateOf<Job?>(null) }
 
     LaunchedEffect(app.wallets) {
         if (draggedWalletId == null) walletList = app.wallets
@@ -185,6 +188,7 @@ fun SidebarView(
                             detectDragGesturesAfterLongPress(
                                 onDragStart = {
                                     haptics.performHapticFeedback(HapticFeedbackType.LongPress)
+                                    preDragSnapshot = walletList
                                     draggedWalletId = wallet.id
                                     draggedOffsetY = 0f
                                 },
@@ -217,20 +221,28 @@ fun SidebarView(
                                 },
                                 onDragEnd = {
                                     val orderedIds = walletList.map { it.id }
-                                    val previous = app.wallets
-                                    persistReorder(
-                                        scope = scope,
-                                        app = app,
-                                        orderedIds = orderedIds,
-                                        previous = previous,
-                                        haptics = haptics,
-                                        onRollback = { walletList = previous },
-                                    )
+                                    pendingReorderJob =
+                                        persistReorder(
+                                            scope = scope,
+                                            app = app,
+                                            previousJob = pendingReorderJob,
+                                            orderedIds = orderedIds,
+                                            haptics = haptics,
+                                            currentOptimisticIds = { walletList.map { it.id } },
+                                            onAdoptAuthoritative = { authoritative ->
+                                                walletList = authoritative
+                                                app.loadWallets()
+                                            },
+                                            onFailureResync = {
+                                                app.loadWallets()
+                                                walletList = app.wallets
+                                            },
+                                        )
                                     draggedWalletId = null
                                     draggedOffsetY = 0f
                                 },
                                 onDragCancel = {
-                                    walletList = app.wallets
+                                    walletList = preDragSnapshot
                                     draggedWalletId = null
                                     draggedOffsetY = 0f
                                 },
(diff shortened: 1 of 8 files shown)


praveenperera (changes requested):
Serialize Android reorder writes so older drags can't win

If a user performs another reorder before the previous `reorderWallets(...)` call finishes, each drag launches an independent coroutine here and whichever write completes last becomes the persisted order. That means an older drag can overwrite the user's newer order even though the UI already showed the later arrangement. iOS already avoids this by chaining `reorderTask`, and Android needs the same serialization or version check.

Sandipmandal25:
Hey @diyaayay i have just a thought maybe i am wrong you can correct me but can we do the commit history bit more clean? like for example some commits are patching earlier commits in the same pr and the binding regeneration commits are floating separately when they should be squashed into the feature commits that triggered them. can we make the repo clean using 3 4 logical commits. What do you think? 
cc @praveenperera

praveenperera:
@Sandipmandal25 its alright, i'm just going to do a squash merge, I do prefer focused logical commits, but doing a squash merge is easier than trying to enforce this on each PR. Also having these messy commits make it easier to see each change that was made after a review.

Sandipmandal25:
> @Sandipmandal25 its alright, i'm just going to do a squash merge

makes sense thanks!

diyaayay:
> Serialize Android reorder writes so older drags can't win
> 
> If a user performs another reorder before the previous `reorderWallets(...)` call finishes, each drag launches an independent coroutine here and whichever write completes last becomes the persisted order. That means an older drag can overwrite the user's newer order even though the UI already showed the later arrangement. iOS already avoids this by chaining `reorderTask`, and Android needs the same serialization or version check.

Persists are now serialized under a `Mutex` and a post-FFI id-equality guard skips stale success/rollback; on success the UI adopts the authoritative `List<WalletMetadata>` from Rust. Android parity with iOS's `reorderTask` chain.

android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
```diff
@@ -30,15 +31,37 @@ import androidx.compose.material3.Icon
...
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import androidx.compose.ui.zIndex
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
```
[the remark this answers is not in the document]
diyaayay: Keeping the IO dispatch. `reorderWallets` is a synchronous UniFFI call that performs a redb write transaction (blocking disk I/O), and `rememberCoroutineScope()` runs on Dispatchers.Main. iOS does the equivalent with Task.detached. Non-actionable.

ios/Cove/Views/SidebarView.swift
```diff
@@ -149,6 +176,48 @@ struct SidebarView: View {
...
+                        appRef.wallets = persisted
+                    }
+                    HapticFeedback.success.trigger()
+                }
+            } catch {
+                Log.error("Failed to reorder wallets: \(error)")
+                await MainActor.run {
+                    if appRef.wallets.map(\.id) == orderedIds {
+                        appRef.wallets = previous
+                    }
```
praveenperera: `previous` is captured from `app.wallets` before each optimistic reorder, so after one queued reorder fails stale and a later queued reorder also fails, this catch can restore an order that never made it to the database. That leaves the sidebar showing a wallet order that will revert on the next reload. 

Please resync from the authoritative store on failure (`appRef.loadWallets()` or `database.wallets().all()`) instead of assigning `previous`, and move the success haptic inside the same `appRef.wallets.map(\.id) == orderedIds` guard so stale completions do not emit a success signal.

android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
```diff
@@ -199,18 +320,78 @@ fun SidebarView(
...
+            val stillCurrent = currentOptimisticIds() == orderedIds
+
+            result.onSuccess { authoritative ->
+                if (stillCurrent) {
+                    onAdoptAuthoritative(authoritative)
+                    haptics.performHapticFeedback(HapticFeedbackType.TextHandleMove)
+                }
+            }.onFailure { error ->
+                Log.e(TAG, "Failed to reorder wallets: ${error.message}")
+                if (stillCurrent) onRollback()
```
praveenperera: This still has the same stale-rollback hole as the iOS path: `onRollback` restores `preDragSnapshot`, but that snapshot can already be an optimistic order from an earlier queued drag. If two queued persists fail back to back, the second failure leaves `walletList` on an order that never reached Rust. When `stillCurrent` is true on failure, reload the authoritative order from `app.loadWallets()` or `app.database.wallets().all()` instead of restoring the snapshot.

praveenperera (changes requested):
I think the `Mutex` is functionally correct, but I’d prefer sequencing these writes by chaining off the previous reorder `Job` instead. What we’re protecting here isn’t really shared mutable state, it’s the ordering of side effects: we need reorder persists to run strictly in drag order so an older drag can’t finish later and write a stale wallet order back over a newer one.

Using a `pendingReorderJob?.join()` style flow would make that intent more explicit, and it would also bring Android in line with iOS, which is already doing the equivalent by awaiting the previous `Task` before starting the next persist. Keeping both platforms on the same serialization pattern will make this code easier to reason about, easier to review, and less likely to drift into subtle platform-specific behavior the next time we touch wallet reordering.

android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
diyaayay: On failure we now resync via `app.loadWallets()` + `walletList = app.wallets` instead of restoring `preDragSnapshot`.

The explicit `walletList` reassignment is needed because Compose's `mutableStateOf` uses structural equality. If the DB still holds the pre-drag order, `LaunchedEffect(app.wallets)` wouldn't re-fire on its own and `walletList` would stay on the failed optimistic order.

```kotlin
onFailureResync = {
    app.loadWallets()
    walletList = app.wallets
}
```

`preDragSnapshot` is retained only for `onDragCancel` (gesture canceled before a persist is issued), which is a pure pre-drag UI restore and doesn't have the staleness risk.

ios/Cove/Views/SidebarView.swift
diyaayay: Success haptic is now inside the `stillCurrent` guard, and on failure we `appRef.loadWallets()` instead of assigning `previous` — the `previous` parameter is gone entirely so there should be no stale UI snapshot left on the rollback path.

android/app/src/main/java/org/bitcoinppl/cove/sidebar/SidebarView.kt
```diff
@@ -47,12 +69,35 @@ import org.bitcoinppl.cove_core.RouteFactory
...
+    val density = LocalDensity.current
+    val haptics = LocalHapticFeedback.current
+    val scope = rememberCoroutineScope()
+    val spacingPx = with(density) { INTER_ITEM_SPACING_DP.dp.toPx() }
+
+    // local copy lets drags apply optimistically; resyncs from app.wallets when idle
+    var walletList by remember { mutableStateOf(app.wallets) }
+    var draggedWalletId by remember { mutableStateOf<WalletId?>(null) }
+    var draggedOffsetY by remember { mutableFloatStateOf(0f) }
+    var itemHeightPx by remember { mutableFloatStateOf(0f) }
```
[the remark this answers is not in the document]
diyaayay: Declining — wallet rows are uniform by design, so `step = itemHeightPx + spacingPx` is correct. Font-scale changes are already handled since `itemHeightPx` updates on every measure. If rows ever become non-uniform, drag-reorder needs a holistic rethink, not just per-wallet height tracking.

diyaayay:
> I think the `Mutex` is functionally correct, but I’d prefer sequencing these writes by chaining off the previous reorder `Job` instead. What we’re protecting here isn’t really shared mutable state, it’s the ordering of side effects: we need reorder persists to run strictly in drag order so an older drag can’t finish later and write a stale wallet order back over a newer one.
> 
> Using a `pendingReorderJob?.join()` style flow would make that intent more explicit, and it would also bring Android in line with iOS, which is already doing the equivalent by awaiting the previous `Task` before starting the next persist. Keeping both platforms on the same serialization pattern will make this code easier to reason about, easier to review, and less likely to drift into subtle platform-specific behavior the next time we touch wallet reordering.

@praveenperera Switched to `pendingReorderJob?.join()` chaining, structurally equivalent to iOS's `await pending?.value`. Also resync from the DB on failure on both platforms now, and moved the iOS success haptic inside the `stillCurrent` guard. Let me know if any other changes are needed.

praveenperera:
thanks @diyaayay this looks good, but I need to do some manual testing on device, will do that and merge soon. Thanks for your contribution.

praveenperera:
Closing this for now after maintainer triage. The wallet reorder branch is conflicting and still has persistence/migration concerns, including ignoring migration failure for the new durable ordering invariant. Please send an updated PR after rebasing and tightening the migration/reconcile behavior.

Issue #295, opened 2025-06-10
When a WalletManager is dropped in AppManager, stop background tasks


praveenperera:
Current status (code inspection on April 16, 2026): this still appears to be an issue.

## What is still wrong

- `SelectedWalletDisappeared` does not actually stop wallet scan work in Rust. It routes to `WalletActor::stop_all_scans()`, but that currently only clears transaction watchers and still has a TODO for stopping wallet scans.
- iOS still keeps `WalletManager` alive in screen-local state/environment instead of making `AppManager` the clear owner. Representative paths:
  - `ios/Cove/Flows/SelectedWalletFlow/SelectedWalletContainer.swift`
  - `ios/Cove/Flows/SelectedWalletFlow/SelectedWalletScreen.swift`
  - `ios/Cove/Flows/SendFlow/SendFlowContainer.swift`
- `ios/Cove/AppManager.swift` still replaces the cached `walletManager` by assigning a new one, without an explicit release/cleanup step for the previous active manager.
- `SelectedWalletContainer.loadManager()` still has a guard that preserves its local `manager` even when `app.walletManager` has been cleared, which weakens the `ClearCachedWalletManager` path.

## Why this matters

- Switching wallets or resetting routes can leave scan work running for a wallet that is no longer active.
- Reconcile messages can still target a manager that is no longer the active manager cached by `AppManager`.
- On iOS there is still not one clear owner for wallet-manager lifetime, so cleanup depends on view teardown / ARC timing instead of an explicit lifecycle boundary.

## Suggested repro

1. Open wallet A so the initial / expanded scan starts.
2. Before the scan finishes, switch to wallet B or trigger a route reset that dismisses the selected-wallet flow.
3. The disappearing screen sends `SelectedWalletDisappeared`, but the Rust cleanup path only clears transaction watchers and does not cancel wallet scans.
4. On iOS, the old `WalletManager` can also remain alive through screen state / environment until ARC tears it down.

## Expected behavior

- When the selected wallet view disappears or `AppManager` swaps wallets, all background work owned by the old `WalletManager` should stop promptly.
- `AppManager` should be the clear owner of the active wallet manager, and screens should reacquire it from `AppManager` instead of holding their own long-lived copies.

## Likely fix direction

- Store abort/cancellation handles for initial, expanded, and incremental scan futures in `WalletActor`, and cancel them from `SelectedWalletDisappeared` and/or `Drop`.
[...]

PR #699, opened 2026-04-22
stop wallet scans when wallet disappears

Sandipmandal25:
## Summary

`stop_all_scans` had a TODO for stopping wallet scans but was only clearing transaction watchers. Scan futures were being spawned without storing their handles, so they could not be cancelled.

This change stores an abort handle for each scan (initial, expanded, incremental) when it starts, and cancels them in `stop_all_scans` and `Drop`. This ensures scans stop promptly when the wallet disappears or the actor is dropped.

fixes part of #295 

## Checklist

### DIFF
--- a/rust/src/manager/wallet_manager/actor.rs
+++ b/rust/src/manager/wallet_manager/actor.rs
@@ -1162,6 +1162,8 @@ impl WalletActor {
         debug!("perform_initial_full_scan");
         static FULL_SCAN_TYPE: FullScanType = FullScanType::Initial;
 
+        self.scan_generation += 1;
+
         let (full_scan_request, graph, node_client) = self.get_for_full_scan().await?;
 
         let addr = self.addr.clone();
@@ -1176,13 +1178,15 @@ impl WalletActor {
             let now = UNIX_EPOCH.elapsed().unwrap().as_secs();
             debug!("[initial] done initial full scan in {}s", now - start);
 
-            // update wallet state
-            let _ =
+            // update wallet state, schedule expanded scan only if initial result was applied
+            let should_schedule_expanded =
                 call!(addr.handle_full_scan_complete(full_scan_result, FULL_SCAN_TYPE, generation))
-                    .await;
+                    .await
+                    .unwrap_or(false);
 
-            // perform next scan
-            send!(addr.maybe_perform_expanded_full_scan());
+            if should_schedule_expanded {
+                send!(addr.maybe_perform_expanded_full_scan());
+            }
         });
         if let Some(old) = self.scan_task.replace(handle.abort_handle()) {
             old.abort();
@@ -1233,6 +1237,7 @@ impl WalletActor {
     async fn perform_incremental_scan(&mut self) -> ActorResult<()> {
         debug!("starting incremental scan");
 
+        self.scan_generation += 1;
         self.state = ActorState::PerformingIncrementalScan;
         let start = UNIX_EPOCH.elapsed().unwrap().as_secs();
 
@@ -1265,10 +1270,13 @@ impl WalletActor {
         full_scan_result: Result<FullScanResponse<KeychainKind>, crate::node::client::Error>,
         full_scan_type: FullScanType,
         generation: u64,
-    ) -> ActorResult<()> {
+    ) -> ActorResult<bool> {
         if generation != self.scan_generation {
-            debug!("dropping stale full scan result (gen {generation} != {})", self.scan_generation);
-            return Produces::ok(());
+            debug!(
+                "dropping stale full scan result (gen {generation} != {})",
+                self.scan_generation
+            );
+            return Produces::ok(false);
         }
 
         debug!("applying full scan result for {full_scan_type:?}");
@@ -1298,7 +1306,7 @@ impl WalletActor {
         // update the state
         self.state = ActorState::FullScanComplete(full_scan_type);
 
-        Produces::ok(())
+        Produces::ok(matches!(full_scan_type, FullScanType::Initial))
     }
 
     async fn handle_incremental_scan_complete(
--- a/rust/src/manager/wallet_manager/actor.rs
+++ b/rust/src/manager/wallet_manager/actor.rs
@@ -72,6 +72,7 @@ pub struct WalletActor {
     seed: u64,
     transaction_watchers: HashMap<Txid, Addr<TransactionWatcher>>,
     scan_task: Option<tokio::task::AbortHandle>,
+    scan_generation: u64,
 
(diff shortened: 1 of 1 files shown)


Sandipmandal25:
fixed the rust side `stop_all_scans` now cancels the active scan task. android side is next will do that in next pr.
cc @praveenperera

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -1190,6 +1195,7 @@ impl WalletActor {
             // update wallet state
             send!(addr.handle_full_scan_complete(full_scan_result, FULL_SCAN_TYPE));
         });
+        self.scan_task = Some(handle.abort_handle());
 
         Produces::ok(())
```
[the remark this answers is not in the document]
Sandipmandal25: scans run sequentially so the old handle is always for a finished task, .abort() on a finished task is a no op in tokio so this is harmless. done the change anyway as its defensive

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -1150,7 +1155,7 @@ impl WalletActor {
         let (full_scan_request, graph, node_client) = self.get_for_full_scan().await?;
 
         let addr = self.addr.clone();
-        self.addr.send_fut(async move {
+        let handle = tokio::spawn(async move {
```
praveenperera: use cove_tokio::task

praveenperera (changes requested):

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: resolved used standard way first missed this! thanks.

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -937,8 +939,11 @@ impl WalletActor {
 
     pub async fn stop_all_scans(&mut self) {
         debug!("stop_all_scans");
+        if let Some(handle) = self.scan_task.take() {
```
praveenperera, line 944: Aborting the stored handle only cancels the scan future while it is still polling. Once the scan task has returned from the node RPC and enqueued `handle_full_scan_complete` or `handle_incremental_scan_complete`, this handle may already be finished, so aborting it during `SelectedWalletDisappeared` is a no-op and the queued completion can still apply wallet updates, persist metadata, and emit UI messages for a wallet the UI has left. Please carry a cancellation generation/session id through each scan completion, invalidate it in `stop_all_scans`, and have completion handlers drop stale results before mutating wallet state.

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -937,8 +939,11 @@ impl WalletActor {
...
+            handle.abort();
+        }
+        self.state = ActorState::Initial;
         self.transaction_watchers = HashMap::default();
```
praveenperera, line 949: Clearing `transaction_watchers` still leaves the delayed `send_fut` scheduled by `mark_transaction_found` alive. If a transaction is found and the selected wallet disappears before that 30-second sleep finishes, the delayed closure will still call `perform_scan_for_single_tx_id`, and that scan is not stored in `scan_task`, so the old wallet can keep scanning and emitting transaction updates after cleanup. Track and abort that delayed task, or gate `perform_scan_for_single_tx_id` with the same cancellation state used by `stop_all_scans`.

praveenperera (changes requested):

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: done added a generation counter that increments in stop_all_scans, each completion handler drops stale results before touching wallet state

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: gated the single tx scan by checking the watcher map at entry since stop_all_scans clears it the 30s delayed rescan becomes a no op after stop

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -1161,11 +1177,16 @@ impl WalletActor {
             debug!("[initial] done initial full scan in {}s", now - start);
 
             // update wallet state
-            let _ = call!(addr.handle_full_scan_complete(full_scan_result, FULL_SCAN_TYPE)).await;
+            let _ =
+                call!(addr.handle_full_scan_complete(full_scan_result, FULL_SCAN_TYPE, generation))
+                    .await;
 
             // perform next scan
             send!(addr.maybe_perform_expanded_full_scan());
```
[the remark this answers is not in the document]
praveenperera: @Sandipmandal25 ^

praveenperera: This still schedules the expanded full scan even when `handle_full_scan_complete` dropped the initial scan result as stale. After `stop_all_scans` increments `scan_generation` and resets `state` to `Initial`, an already-returned initial scan task can reach this line, enqueue `maybe_perform_expanded_full_scan`, and start a fresh expanded scan for the disappeared wallet using the new generation. Make the completion handler report whether the result was accepted, or re-check the generation before this send, and only schedule the expanded scan when the initial completion was current.

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -1217,16 +1242,20 @@ impl WalletActor {
         let graph = self.wallet.bdk.tx_graph().clone();
 
         let addr = self.addr.clone();
-        self.addr.send_fut(async move {
+        let generation = self.scan_generation;
```
praveenperera, line 1250: New tracked scan launches reuse the existing `scan_generation`, so aborting a previous handle does not invalidate a completion that was already enqueued before the abort. In that race, the old completion carries the same generation as the replacement scan and still passes the stale-result check. Mint a new scan session/generation when starting a new top-level scan, while preserving the same session across the intended initial-to-expanded full-scan chain.

praveenperera (changes requested):

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: Moving the expanded scan trigger into the completion handler so it only fires when the result was accepted.

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: Will mint a fresh generation at the start of each top level scan launch. Keeping expanded on the same generation as initial since they re one chain.

Sandipmandal25:
i am making the chnages and updating the pr shortly

praveenperera:
Thanks @Sandipmandal25  merged

Issue #491, opened 2025-12-31
Add watch-only import for multi-sig public descriptors


praveenperera:
### Summary

Allow users to import multi-sig public descriptors as watch-only wallets. This enables users to monitor their multi-sig wallet balances in Cove without any signing or broadcasting capability.

### Context

A user requested the ability to "watch" a multisig wallet in Cove - explicitly with no ability to sign and broadcast transactions, genuinely only watching the wallet balance and transactions.

### Scope

**What this is:**
- Import multi-sig public descriptors (e.g., `wsh(sortedmulti(...))`) as watch-only
- View balance and transaction history
- Same experience as current watch-only single-sig wallets

**What this is NOT:**
- Full multisig support (signing, PSBT coordination, etc.)
- Any transaction creation or broadcasting capability

### Why watch-only only?

Full multisig support is explicitly out of scope for Cove. There are already excellent wallets for multisig (Nunchuk) and Lightning (Phoenix, Bull Bitcoin). Adding signing/coordination features would complicate Cove and compromise its goal of being a **simple** Bitcoin wallet.

Watch-only import allows users who manage their multisig elsewhere to conveniently check balances in Cove alongside their other wallets.

PR #702, opened 2026-04-23
feat: watch-only import for wsh(sortedmulti) descriptors

DhruvArvindSingh:
## Summary

Implements watch-only multisig wallet import via wsh(sortedmulti) descriptors, as scoped in issue #491.

## Testing

- Wallet::new_from_multisig_descriptor : creates and persists the BDK wallet                
- RustWalletManager::try_new_from_multisig_descriptor : UniFFI-exported entry point, mirrors the existing try_new_from_xpub pattern

The parse_multisig_descriptors helper accepts two formats:
- Two newline-separated descriptors (external then internal), as exported by most hardware coordinators (Sparrow, Nunchuk, etc.)
- A single BIP-389 multipath descriptor using <0;1> notation

I have't tested it, as their is no UI wired up yet

### Platform Coverage

- [ ] Tested on iOS device
- [ ] Tested on iOS simulator
- [x] Not tested

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for creating watch-only wallets directly from multisig descriptors
  * Implemented automatic duplicate wallet detection and prevention to avoid redundant configurations
  * Extended wallet initialization APIs with new descriptor-based constructor methods

* **Tests**
  * Added comprehensive unit tests validating multisig descriptor parsing and wallet creation flows

### DIFF
--- a/rust/src/manager/wallet_manager.rs
+++ b/rust/src/manager/wallet_manager.rs
@@ -388,6 +388,34 @@ impl RustWalletManager {
         })
     }
 
+    #[uniffi::constructor]
+    pub fn try_new_from_multisig_descriptor(descriptor: String) -> Result<Self, Error> {
+        let (sender, receiver) = flume::bounded(100);
+
+        let wallet = Wallet::try_new_persisted_multisig_watch_only(descriptor)?;
+        let id = wallet.id.clone();
+        let metadata = wallet.metadata.clone();
+
+        let scanner =
+            WalletScanner::try_new(metadata.clone(), sender.clone()).ok().map(spawn_actor);
+
+        let wallet_actor = WalletActor::new(wallet, sender.clone())
+            .map_err(|e| Error::DatabaseCorruption { id: id.clone(), error: e.to_string() })?;
+        let actor = task::spawn_actor(wallet_actor);
+        let label_manager = LabelManager::new(id.clone()).into();
+
+        Ok(Self {
+            id,
+            actor,
+            metadata: Arc::new(RwLock::new(metadata)),
+            reconciler: MessageSender::new(sender),
+            reconcile_receiver: Arc::new(receiver),
+            label_manager,
+            initial_load_state: WalletLoadState::Loading,
+            scanner,
+        })
+    }
+
     #[uniffi::constructor(default(backup = None))]
     pub fn try_new_from_tap_signer(
         tap_signer: Arc<cove_tap_card::TapSigner>,
--- a/rust/src/manager/wallet_manager.rs
+++ b/rust/src/manager/wallet_manager.rs
@@ -396,12 +396,12 @@ impl RustWalletManager {
         let id = wallet.id.clone();
         let metadata = wallet.metadata.clone();
 
-        let scanner =
-            WalletScanner::try_new(metadata.clone(), sender.clone()).ok().map(spawn_actor);
-
         let wallet_actor = WalletActor::new(wallet, sender.clone())
             .map_err(|e| Error::DatabaseCorruption { id: id.clone(), error: e.to_string() })?;
         let actor = task::spawn_actor(wallet_actor);
+
+        let scanner =
+            WalletScanner::try_new(metadata.clone(), sender.clone()).ok().map(spawn_actor);
         let label_manager = LabelManager::new(id.clone()).into();
 
         Ok(Self {
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -374,10 +374,10 @@ impl Wallet {
         let external_str = external.to_string();
         let existing = database.wallets.get_all(network, mode).unwrap_or_default();
         for wm in existing {
-            if let Ok(Some((stored_ext, _))) = keychain.get_public_descriptor(&wm.id) {
-                if stored_ext.to_string() == external_str {
-                    return Err(WalletError::WalletAlreadyExists(wm.id));
-                }
+            if let Ok(Some((stored_ext, _))) = keychain.get_public_descriptor(&wm.id)
+                && stored_ext.to_string() == external_str
+            {
+                return Err(WalletError::WalletAlreadyExists(wm.id));
             }
         }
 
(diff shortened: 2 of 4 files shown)


Vineet1101:
@DhruvArvindSingh I think this is related to the project work and it should be done by the selected contributor for the project. What are your thoughts on this @praveenperera

rust/src/wallet.rs
```diff
@@ -361,6 +361,48 @@ impl Wallet {
...
+            external: Descriptor { extended_descriptor: external.clone(), key_map: KeyMap::new() },
+            internal: Descriptor { extended_descriptor: internal.clone(), key_map: KeyMap::new() },
+        }
+        .into_create_params()
+        .network(network.into())
+        .create_wallet(&mut store.conn)
+        .map_err_str(WalletError::BdkError)?;
+
+        keychain.save_public_descriptor(&id, external, internal)?;
+        database.wallets.save_new_wallet_metadata(metadata.clone())?;
```
praveenperera, line 400: If `create_wallet` succeeds but `save_public_descriptor` or `save_new_wallet_metadata` fails, this returns an error after creating the BDK store and possibly saving keychain data. That leaves orphaned wallet state behind even though the import is reported as failed. Wrap the create/persist sequence in a fallible block and, on any error after the new `id` is allocated, clean up with `keychain.delete_wallet_items(&id)`, `delete_wallet_specific_data(&id)`, and `database.wallets.delete(&id)` before returning the original error. Only call `CLOUD_BACKUP_MANAGER.handle_wallet_set_change()` after all persistence succeeds.

rust/src/manager/wallet_manager.rs
```diff
@@ -388,6 +388,34 @@ impl RustWalletManager {
...
+    #[uniffi::constructor]
+    pub fn try_new_from_multisig_descriptor(descriptor: String) -> Result<Self, Error> {
+        let (sender, receiver) = flume::bounded(100);
+
+        let wallet = Wallet::try_new_persisted_multisig_watch_only(descriptor)?;
+        let id = wallet.id.clone();
+        let metadata = wallet.metadata.clone();
+
+        let wallet_actor = WalletActor::new(wallet, sender.clone())
+            .map_err(|e| Error::DatabaseCorruption { id: id.clone(), error: e.to_string() })?;
```
praveenperera, line 400: `Wallet::try_new_persisted_multisig_watch_only` has already persisted the wallet before `WalletActor::new` opens the wallet data DB. If `WalletActor::new` fails here, the constructor returns `DatabaseCorruption` while the imported wallet remains in BDK/keychain/database state, so the user sees a failed import but may be blocked from retrying by the duplicate check. Handle the `WalletActor::new` error explicitly and delete the just-created wallet data/keychain items/metadata before returning.

rust/src/wallet.rs
```diff
@@ -712,10 +754,119 @@ pub fn delete_wallet_specific_data(wallet_id: &WalletId) -> eyre::Result<()> {
...
+            let int_str = single.replace("<0;1>", "1");
+            (parse(&ext_str)?, parse(&int_str)?)
+        }
+        _ => return Err(WalletError::UnsupportedWallet(
+            "expected two descriptor lines (external, internal) or a single multipath descriptor"
+                .to_string(),
+        )),
+    };
+
+    let name = match &external {
```
praveenperera, line 801: Only the external descriptor is validated as `wsh(sortedmulti(...))`; the internal descriptor can be any parseable descriptor in the two-line format. That can persist a wallet whose change descriptor is unsupported or mismatched while still naming it from the external multisig shape. Validate `internal` with the same `WshInner::SortedMulti` check and reject mismatched threshold/signer count before returning the descriptors.

rust/src/wallet.rs
```diff
@@ -712,10 +754,119 @@ pub fn delete_wallet_specific_data(wallet_id: &WalletId) -> eyre::Result<()> {
...
+            .map(|(d, _)| d)
+            .map_err_str(WalletError::DescriptorKeyParseError)
+    };
+
+    let lines: Vec<&str> = input.lines().map(str::trim).filter(|l| !l.is_empty()).collect();
+
+    let (external, internal) = match lines.as_slice() {
+        [ext, int] => (parse(ext)?, parse(int)?),
+        [single] if single.contains("<0;1>") => {
+            let ext_str = single.replace("<0;1>", "0");
```
praveenperera, line 791: This expands `<0;1>` before parsing, but checksummed descriptors from common exports include a trailing `#checksum`. Replacing the multipath segment changes the descriptor body while leaving the old checksum attached, so valid checksummed multipath imports are rejected by descriptor parsing. Strip any `#...` suffix before building `ext_str` and `int_str`, then parse the checksum-free expanded descriptors. Add some failing tests that confirm this before fixing.

rust/src/wallet.rs
```diff
@@ -361,6 +361,48 @@ impl Wallet {
...
+
+        let (external, internal, name) = parse_multisig_descriptors(&descriptor)?;
+
+        let keychain = Keychain::global();
+        let database = Database::global();
+        let network = database.global_config.selected_network();
+        let mode = database.global_config.wallet_mode();
+
+        let external_str = external.to_string();
+        let existing = database.wallets.get_all(network, mode).unwrap_or_default();
```
praveenperera, line 375: The duplicate check silently falls back to an empty wallet list when `get_all` fails, and line 377 also ignores keychain read/parse errors. A transient DB/keychain error can therefore bypass duplicate detection and create another copy of the same watch-only wallet. Propagate these errors during import instead of treating them as “no duplicate”; compare the parsed stored descriptor to the parsed new external descriptor rather than relying on string rendering.

praveenperera (changes requested):

praveenperera:
Closing this for now after maintainer triage. The watch-only descriptor import branch is conflicting and still has import correctness issues around persistence rollback, duplicate detection, checksummed multipath descriptors, and internal descriptor validation.

PR #706, opened 2026-04-26
Feature add new badge for first-time addresses

### LINKED ISSUE #227: Add `new` badge if sending to an address you've never sent to before

oyetanishq:
## Summary
- Show `New Address` badge when sending to an address you've never sent to before.
- Close: #227 

## Testing
|IOS|Android|
|-|-|
|Screenshot 2026-04-26 at 16 02 41|Screenshot 2026-04-26 at 17 51 30|

### Platform Coverage

- [x] Tested on iOS device
- [x] Tested on iOS simulator
- [x] Tested on Android simulator
- [ ] Not tested

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a "New Address" badge shown in address entry and confirmation when the recipient address appears unused.
  * Badge state is determined automatically from wallet history and updates as you type or scan an address.
  * The badge is cleared when the address field is cleared.

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/EnterAddressView.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/EnterAddressView.kt
@@ -8,10 +8,15 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.text.BasicTextField
 import androidx.compose.foundation.text.KeyboardActions
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.KeyboardOptions
+import org.bitcoinppl.cove.ui.theme.CoveColor
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Clear
 import androidx.compose.material.icons.filled.QrCode2
@@ -43,6 +48,7 @@ import org.bitcoinppl.cove_core.types.addressStringSpacedOut
 fun EnterAddressView(
     onScanQr: () -> Unit,
     initialAddress: String,
+    isNewAddress: Boolean = false,
     onAddressChanged: (String) -> Unit,
     focusRequester: FocusRequester,
     onFocusChanged: (Boolean) -> Unit = {},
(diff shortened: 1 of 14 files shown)


rust/src/manager/send_flow_manager.rs
```diff
@@ -891,7 +893,18 @@ impl RustSendFlowManager {
...
         let me = self.clone();
         let is_max_selected = self.state.lock().max_selected.is_some();
+        let valid_address = address.clone();
         cove_tokio::task::spawn(async move {
+            if let Some(address) = valid_address {
+                let actor = me.wallet_actor();
+                let cloned_address = (*address).clone();
+
+                let has_sent =
+                    call!(actor.has_sent_to_address(cloned_address)).await.unwrap_or(false);
```
harsh04044: `unwrap_or(false)` means if the actor call fails, `has_sent` is false, so `!has_sent` is true, and the "New Address" badge shows for every address on error. Safer default would be `unwrap_or(true)` so the badge doesn't show on failure.

rust/src/manager/send_flow_manager.rs
```diff
@@ -891,7 +893,18 @@ impl RustSendFlowManager {
         // when we have a valid address, use that to get the fee rate options
         let me = self.clone();
         let is_max_selected = self.state.lock().max_selected.is_some();
+        let valid_address = address.clone();
         cove_tokio::task::spawn(async move {
+            if let Some(address) = valid_address {
```
harsh04044, line 898: when `valid_address` is `None` (user edits the address into something invalid), `UpdateIsNewAddress` is never sent, so the previous `true` lingers. Adding an `else` branch that sends `UpdateIsNewAddress(false)` would fix it.

harsh04044 (changes requested):

rust/src/manager/send_flow_manager.rs
oyetanishq: skipped my mind, I was returning `!has_sent` 😅

rust/src/manager/send_flow_manager.rs
```diff
@@ -891,7 +893,21 @@ impl RustSendFlowManager {
...
+                let actor = me.wallet_actor();
+                let cloned_address = (*address).clone();
+
+                let has_sent =
+                    call!(actor.has_sent_to_address(cloned_address)).await.unwrap_or(true);
+
+                me.reconciler.send(Message::UpdateIsNewAddress(!has_sent));
```
praveenperera, line 905: Each address edit spawns an async task and sends `UpdateIsNewAddress` when the wallet lookup finishes, but the result is not checked against the address that is still current. If a user enters valid address A and quickly changes to address B or an invalid value, A's delayed lookup can arrive last and show or hide the badge for the wrong recipient. Before sending this reconcile message, compare the captured address with `state.address` (or use a generation token/cancel previous lookup) and discard stale results.

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/EnterAddressView.kt
```diff
@@ -67,12 +73,26 @@ fun EnterAddressView(
...
+                    Text(
+                        stringResource(R.string.label_enter_address),
+                        color = MaterialTheme.colorScheme.onSurface,
+                        fontSize = 18.sp,
+                        fontWeight = FontWeight.SemiBold,
+                    )
+                    if (isNewAddress) {
+                        Spacer(Modifier.width(8.dp))
+                        Text(
+                            text = "New Address",
```
praveenperera, line 86: The badge text is still hardcoded, and the confirm screen repeats the same literal. That leaves this user-visible label outside Android localization while the surrounding strings use `stringResource(...)`. Add a string resource such as `label_new_address_badge` and use it in both badge call sites, preferably through a shared `NewAddressBadge` composable so the text and styling stay in sync.

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowConfirmScreen.kt
```diff
@@ -442,13 +446,26 @@ private fun KeyValueRow(
...
+                key,
+                color = keyColor,
+                fontSize = 14.sp,
+                fontWeight = if (boldKey) FontWeight.SemiBold else FontWeight.Normal,
+            )
+            if (showNewAddressBadge) {
+                Spacer(Modifier.width(6.dp))
+                Text(
+                    text = "New Address",
+                    color = CoveColor.WarningOrange,
```
praveenperera, line 460: `CoveColor.WarningOrange` on a 20% orange background has very low contrast on the light `MaterialTheme.colorScheme.surface`, especially at `10.sp`, so the badge can be hard to read and fails normal small-text accessibility contrast. Use a darker/theme-aware text color or a higher-contrast badge background, and apply the same badge styling wherever the new-address badge is rendered.

praveenperera (changes requested):

praveenperera:
also need to resolve conflicts on main @oyetanishq

praveenperera:
Closing this for now after maintainer triage. The new-address badge branch is conflicting and still has behavioral risks: async lookup results can race, prefilled addresses can miss lookup, outgoing change outputs can suppress the badge incorrectly, and Android copy/contrast need cleanup.

PR #725, opened 2026-05-02
Improve unsupported SLIP-132 xpub import error

### LINKED ISSUE #723: Improve error message for unsupported zpub (SLIP-132) when importing hardware wallet
### Problem
When trying to import a hardware wallet using a **zpub** (the default SLIP-132 format exported by most hardware wallets like Coldcard, Ledger, Trezor, etc.), the app shows this technical error:

>CoveCore.WalletError.ParseXpubError (CoveCore.XpubError.InvalidXpub(“Invalid xpub: unknown version magic bytes: [4, 178, 71, 70]”))

Most users have no idea what “magic bytes” means, so they get confused and don’t know what to do next.

### Suggestion (UX improvement only)
Improve the error message to be user-friendly and actionable, something like:

> **zpub format is not supported**  
> Cove currently accepts **xpub** (BIP32) or **public descriptors**.  

This would make the onboarding experience much smoother for hardware wallet users without requiring any code changes to parsing logic.

### Why this matters
- The app already documents support for “xpub / public descriptor” in the features, but the error doesn’t guide the user.
- Many people hit this exact issue when following the most common hardware wallet export flow.
- I understand that adding full zpub parsing isn’t feasible because of BDK. This is purely about better error messaging.

Thanks for the great wallet!

kanishka0411:
## Summary
Fixes #723 

This improves the error returned when a bare SLIP-132 public key, like `zpub`, is pasted into the direct xpub import path.

Cove still accepts BIP32 `xpub`/`tpub` keys and public descriptors here. This does not add `zpub` import support or change accepted import formats; it only avoids leaking the low-level BIP32 “magic bytes” parser error and gives the user an actionable message instead.

## Changes
- Detect common SLIP-132 public key prefixes before direct xpub parsing
- Reuse the existing `XpubError::InvalidXpub` path to avoid UniFFI binding churn
- Add a regression test for direct `zpub` import

## Testing
- `cargo fmt --all`
- `just clippy`
- `just lint-rust`
- `just ctest importing_slip132_xpub_returns_actionable_error`
- `just test-gh`

### Platform Coverage

- [ ] Tested on iOS device
- [ ] Tested on iOS simulator
- [x] Not tested

## Checklist

## Release Notes

* **Bug Fixes**
  * Improved error handling when importing extended public keys with unsupported formats. The system now detects incompatible key types early and provides clearer, more actionable error messages guiding users toward compatible formats.

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -255,6 +255,10 @@ impl Wallet {
     /// Create a new watch-only wallet from the given xpub
     pub fn try_new_persisted_from_xpub(xpub: String) -> Result<Self, WalletError> {
         let xpub = xpub.trim();
+        if let Some(error) = xpub::XpubError::unsupported_slip132_prefix(xpub) {
+            return Err(WalletError::ParseXpubError(error));
+        }
+
         let hardware_export = pubport::Format::try_new_from_str(xpub)
             .map_err(Into::into)
             .map_err(WalletError::ParseXpubError);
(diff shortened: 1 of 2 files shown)


praveenperera:
This is a good small change, but I would rather support these formats long term @kanishka0411

praveenperera:
@kanishka0411 actually i'm closing this, better to handle this in the `pubport` crate

PR #752, opened 2026-05-15
Wire PayJoin Endpoint Through Confirm Send Flow

Sandipmandal25:
## Summary

PR #738 stored the PayJoin endpoint in `SendFlowManagerState`, but the confirm screen was still not receiving it and continued using the normal `signAndBroadcastTransaction` path.

This PR wires the endpoint through the send flow by adding `payjoin_endpoint` to `SendRouteConfirmArgs`, forwarding it into the confirm screen, and updating iOS and Android to call `initiate_payment` / `initiatePayment` for unsigned sends.

For now, the new `initiate_payment` entry point is stubbed. It accepts the PayJoin endpoint but intentionally delegates to the existing sign-and-broadcast flow, so runtime behavior stays unchanged.

The actual PayJoin negotiation logic will be added in Milestone 3.

## Testing

No user-facing behavior change is expected.

- Sends without `pj=` work as before
- Sends with `pj=` also work as before for now
- Existing `cove` tests pass

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/bitcoinppl/cove/blob/master/CONTRIBUTING.md)
- [x] I have read [ARCHITECTURE.md](https://github.com/bitcoinppl/cove/blob/master/ARCHITECTURE.md)

## Summary by CodeRabbit

* **Chores**
  * Enhanced payment confirmation workflow to support optional endpoint parameters
  * Refactored unsigned wallet transaction submission to use an improved payment initiation method
  * Improved backend routing infrastructure for payment confirmation processing across Android and iOS platforms
  * Unified payment state management for better consistency across platforms
  * Updated test configuration to properly isolate network-dependent tests

[Review Change Stack](https://app.coderabbit.ai/change-stack/bitcoinppl/cove/pull/752?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #772, opened 2026-05-28
Implement BIP77 v2 async PayJoin sender

Sandipmandal25:
## Summary

The previous PR (#752) wired the PayJoin endpoint through the confirm screen, but `initiate_payment` still treated it as a stub and ignored the endpoint, falling back to the existing sign-and-broadcast path.

This PR implements the actual BIP77 v2 asynchronous send flow.

When a PayJoin endpoint is present, the original PSBT is signed first to produce both the signed PSBT used for negotiation and a valid fallback transaction. A dedicated `PayjoinActor` then manages the negotiation:

- Posts the signed PSBT to the PayJoin directory through a randomly selected OHTTP relay
- Long-polls the directory for the receiver's proposal, retrying transient client and relay failures with a short delay
- Signs and broadcasts the PayJoin proposal when negotiation succeeds
- Broadcasts the pre-signed fallback transaction when negotiation fails
- Terminates the PayJoin actor after a terminal success or fallback result

The sender disables output substitution when constructing the PayJoin session, preserving `pjos=0`.

Both Android and iOS listen for the `PayjoinTxBroadcast` reconcile message and update the send UI after either the PayJoin proposal or fallback transaction is successfully broadcast. Broadcast failures are propagated through the existing send-flow error state.

Non-PayJoin sends remain unchanged and continue using the original synchronous sign-and-broadcast path.

## Testing

* Verified that non-PayJoin sends with no `pj=` endpoint continue through the existing synchronous path
* Verified the PayJoin success and fallback flows

## Checklist

## Release Notes

* **New Features**
  * Added PayJoin BIP77 v2 support to the send flow. When a PayJoin-enabled endpoint is used, the app negotiates PayJoin and completes the send based on the final broadcast result.
  * Improved PayJoin completion handling across platforms with explicit tracking of each PayJoin broadcast event.

* **Bug Fixes**
  * Fixed unsigned-send behavior to avoid unlocking/showing success immediately during PayJoin; the UI now updates only after the PayJoin broadcast succeeds or fails, showing the correct success/error alerts.
  * Enhanced fallback handling: if PayJoin negotiation fails or broadcast delivery fails, the app broadcasts a standard transaction and unblocks with the appropriate alert.


rust/src/manager/wallet_manager/actor.rs
```diff
@@ -744,6 +793,46 @@ impl WalletActor {
...
+        &mut self,
+        fallback_tx: BdkTransaction,
+    ) -> ActorResult<()> {
+        self.do_broadcast_transaction(fallback_tx.clone())
+            .await
+            .tap_err(|error| error!("payjoin fallback broadcast failed: {error}"))?;
+
+        self.send(WalletManagerReconcileMessage::PayjoinTxBroadcast);
+        Produces::ok(())
+    }
```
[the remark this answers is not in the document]
Sandipmandal25: `do_broadcast_transaction` is generated by `#[into_actor_result]` with the full body of broadcast transaction which already includes insert broadcast transaction. No insert is missing

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -2013,6 +2102,164 @@ impl Drop for WalletActor {
...
+    let encoded_endpoint = endpoint.replace('#', "%23");
+
+    // identify the recipient output: outputs with no bip32_derivation are
+    // external (receiver) outputs — no derivation info in our wallet
+    let (idx, _) = signed_psbt
+        .outputs
+        .iter()
+        .enumerate()
+        .find(|(_, o)| o.bip32_derivation.is_empty())
+        .ok_or_else(|| eyre::eyre!("no recipient output found in PSBT"))?;
```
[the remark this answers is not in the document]
Sandipmandal25: This pr only handles cove's hot wallet single recipient sends. The psbt always has exactly one external output (recipient) and one change output with derivation info so the heuristic is sound for this scope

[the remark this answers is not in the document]
praveenperera: @Sandipmandal25 put a note on when this heuristic would not work, so if we add more features and run into this new situation this comment should catch it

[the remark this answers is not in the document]
Sandipmandal25: added!

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -583,14 +590,47 @@ impl WalletActor {
...
+            match payjoin_http_flow(signed_psbt, endpoint, network).await {
+                Ok(proposal_psbt) => {
+                    send!(addr.handle_payjoin_success(proposal_psbt, fallback_tx))
+                }
+                Err(error) => {
+                    warn!("payjoin negotiation failed, broadcasting fallback tx: {error:?}");
+                    send!(addr.handle_payjoin_fallback(fallback_tx))
+                }
+            }
+        });
```
praveenperera: lets put the payjoin specific code into a different function

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -2039,6 +2144,175 @@ impl Drop for WalletActor {
     }
 }
 
+// max poll attempts before falling back; 60 × 5s = 5 minutes
+const PAYJOIN_MAX_POLL_ATTEMPTS: u32 = 60;
+
+const PAYJOIN_POLL_INTERVAL: Duration = Duration::from_secs(5);
+
```
praveenperera: move the consts and all the rest of the payjoin related functions in this file to a seperate payjoin module

rust/Cargo.toml
```diff
@@ -57,7 +57,7 @@ bitcoin = { version = "0.32.9" }
 bdk_wallet = { version = "3.0.0", features = ["keys-bip39", "file_store", "rusqlite"] }
 bip39 = { version = "2.2.2", features = ["zeroize"] }
 bip329 = { version = "0.4.0" }
-payjoin = { version = "0.25.0", default-features = false, features = ["v1"] }
+payjoin = { version = "0.25.0", default-features = false, features = ["v1", "v2"] }
```
praveenperera: do we still need `v1` feature too?

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
```diff
@@ -423,6 +428,29 @@ private fun SendFlowRouteToScreen(
...
+                if (walletManager.sendFlowErrorAlert != null && sendState == SendState.Sending) {
+                    val message =
+                        when (val alert = walletManager.sendFlowErrorAlert!!.item) {
+                            is SendFlowErrorAlert.SignAndBroadcast -> alert.v1
+                            is SendFlowErrorAlert.ConfirmDetails -> alert.v1
+                        }
+                    sendState = SendState.Error(message)
+                    showErrorAlert = true
+                }
+            }
```
praveenperera, line 484: theres a few launched effects that handle payjoin related functionality as a special case, is this needed or is there a better way to handle this?

praveenperera (changes requested):

rust/Cargo.toml
Sandipmandal25: its used in another crate cove-types uses it for parsing bitcoin URIs.

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
Sandipmandal25: The two effects are needed because payjoin runs in the background the result comes back later via a reconcile message not through the normal send flow. I simplifying it a bit though removing the third reset effect by using the same tagged item pattern we already use for errors.

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -725,6 +774,62 @@ impl WalletActor {
...
+            Err(error) => {
+                error!("payjoin broadcast failed: {error}");
+                self.send(WalletManagerReconcileMessage::SendFlowError(
+                    SendFlowErrorAlert::SignAndBroadcast(error.to_string()),
+                ));
+            }
+        }
+
+        Produces::ok(())
+    }
```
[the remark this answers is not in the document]
Sandipmandal25: its the same false positive flagged by greptile `do_broadcast_transaction` already calls `insert_broadcast_transaction` internally via the `#[into_actor_result] `macro.

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: done

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: yes done

rust/Cargo.toml
praveenperera: @Sandipmandal25 csn we clean it up so we only need v2 feature?

Sandipmandal25: yes can clean it up  the only v1 thing in cove types is the `payjoin::OutputSubstitution` match for pjos. I can replace that with a direct query param check should i do that in a fllow up pr or in this one ?

praveenperera: @Sandipmandal25 follow up PR is fine

rust/src/manager/wallet_manager/payjoin.rs
praveenperera, line 1: this is not an actor, find a better place to put it

praveenperera (changes requested):

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: moved it to wallet manager

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,187 @@
...
+            .save(&persister)
+            .map_err(|e| eyre::eyre!("failed to process poll response: {e:?}"))?
+        {
+            OptionalTransitionOutcome::Progress(proposal_psbt) => return Ok(proposal_psbt),
+            OptionalTransitionOutcome::Stasis(next) => {
+                polling_sender = next;
+                debug!("payjoin poll attempt {attempt}: no proposal yet, continuing");
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Sandipmandal25: agreed its asymmetric `process_response` takes self so retrying isnt straightforward but open to suggestions if theres a way to handle it
cc @praveenperera

rust/Cargo.toml
Sandipmandal25: added a new pr #778

rust/src/manager/wallet_manager/payjoin.rs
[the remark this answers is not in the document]
DanGould: Bakcwards compatible receivers MUST set pjos=0 to prevent PSBT modification by a malicious directory.

https://github.com/bitcoin/bips/blob/master/bip-0077.md#session-initiation

if you're just talking about the sender here, just make sure not to fall back to v1 ever (which is part of the v2 state machine)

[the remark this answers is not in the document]
Sandipmandal25: > Bakcwards compatible receivers MUST set pjos=0 to prevent PSBT modification by a malicious directory.

thanks for the spec ref we were dropping pjos when reconstructing the url in `payjoin_http_flow.` calling `always_disable_output_substitution()` on the builder now.

and i  have a follow up pr that drops the v1 feature from Cargo.toml entirely so it wont even compile after that merges #778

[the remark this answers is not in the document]
praveenperera: @Sandipmandal25 looking at all the negotiation state makes me think we should make a `PayjoinActor` and do all state handling in there, read `ARCHITECTURE.md`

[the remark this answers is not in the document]
Sandipmandal25: i am refactoring into a `PayjoinActor` following the same pattern as `WalletScanActor`

DanGould:
I also see "Poll every 5s up to 5 min," while you'll notice [payjoin-cli imposes no such timeout](https://github.com/payjoin/rust-payjoin/blob/c883af04f11c259484fc4f94712fc40e26a0aaf9/payjoin-cli/src/app/v2/mod.rs#L714-L735). The poll duration is imposed on the directory side https://github.com/payjoin/rust-payjoin/blob/master/payjoin-mailroom/src/config.rs#L89. This is long-polling, so the connection stays open. It won't take 30s for an update to show up in the UI, it'll return as soon as there is one. I'm concerned the 5s polling might fingerprint this client.

5 minute waiting is also something you could opt for, but I would probably recommend against outside of an automated receiver like a store. Instead, I prefer exposing the option for a user to manually abort the payjoin and broadcast the fallback. That's because the receiver actually sets an [expiration](https://github.com/bitcoin/bips/blob/master/bip-0077.md#session-expiration) on the payjoin that they're committed to respond to if they're online. This can be displayed in UI as a reason a session may have not completed. But impatient users can always manually fall back.

There are quite a few of these best practicies baked into payjoin-cli that haven't yet made it into the BIP 77 spec as normative "SHOULD"-dos, but they're coming. We just closed out our 1.0 milestone by releasing [`payjoin-1.0.0-rc.3`](https://crates.io/crates/payjoin/1.0.0-rc.3) (time to audit & review) *minutes ago*. Thanks for getting in here early, I know there's a lot of jockying back and forth when working with experimental pre-stability software. I'm getting a lot of feedback by following this exchange.

Sandipmandal25:
> I also see "Poll every 5s up to 5 min," while you'll notice [payjoin-cli imposes no such timeout](https://github.com/payjoin/rust-payjoin/blob/c883af04f11c259484fc4f94712fc40e26a0aaf9/payjoin-cli/src/app/v2/mod.rs#L714-L735). The poll duration is imposed on the directory side https://github.com/payjoin/rust-payjoin/blob/master/payjoin-mailroom/src/config.rs#L89. This is long-polling, so the connection stays open. It won't take 30s for an update to show up in the UI, it'll return as soon as there is one. I'm concerned the 5s polling might fingerprint this client.
> 
> 5 minute waiting is also something you could opt for, but I would probably recommend against outside of an automated receiver like a store. Instead, I prefer exposing the option for a user to manually abort the payjoin and broadcast the fallback. That's because the receiver actually sets an [expiration](https://github.com/bitcoin/bips/blob/master/bip-0077.md#session-expiration) on the payjoin that they're committed to respond to if they're online. This can be displayed in UI as a reason a session may have not completed. But impatient users can always manually fall back.
> 
> There are quite a few of these best practicies baked into payjoin-cli that haven't yet made it into the BIP 77 spec as normative "SHOULD"-dos, but they're coming. We just closed out our 1.0 milestone by releasing [`payjoin-1.0.0-rc.3`](https://crates.io/crates/payjoin/1.0.0-rc.3) (time to audit & review) _minutes ago_. Thanks for getting in here early, I know there's a lot of jockying back and forth when working with experimental pre-stability software. I'm getting a lot of feedback by following this exchange.

thanks for this ! i am removing the 5s interval and client side timeout entirely. Polls will fire immediately on stasis with the http timeout set to 35s to give the directory's own timeout room to breathe

The manual abort ux is the right call and a good candidate for a followup exposing a cancel button that broadcasts the fallback directly makes more sense than an arbitrary timeout, especially with receiver-set expiration that can be surfaced in the ui

and looking forward to rc.3

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        self.addr.send_fut_with(|addr| async move {
+            let client = match cove_http::new_client() {
+                Ok(c) => c,
+                Err(e) => {
+                    warn!("payjoin poll: failed to create HTTP client: {e:?}");
+                    send!(addr.begin_next_poll_msg());
+                    return;
+                }
+            };
```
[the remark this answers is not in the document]
Sandipmandal25: Intentional client creation is transient. In `begin_poll` we have a valid session worth preserving so retry makes more sense than falling back. start post is different because there's no session yet thats what i think!

praveenperera:
@Sandipmandal25 the actor is auto terminated when the last `Addr<>` is dropped. So best way to terminate is to just the parent actor holding the addr field to none

Sandipmandal25:
> @Sandipmandal25 the actor is auto terminated when the last `Addr<>` is dropped. So best way to terminate is to just the parent actor holding the addr field to none

my thinking was that the watcher closure captures its own strong reference, so setting the parent field to none still leaves one reference alive and i wasnt sure termination would ever resolve with that reference still held  so i went with explicit termination from inside the actor to guarantee it exits. i am fixing it to just clear the parent field in the two result handlers

Sandipmandal25:
the watcher closure also holds a strong addr so when we set the parent field to none the actor still has one ref alive in the watcher. should we remove the watcher too or is there something that handles this.?

praveenperera:
> the watcher closure also holds a strong addr so when we set the parent field to none the actor still has one ref alive in the watcher. should we remove the watcher too or is there something that handles this.?

We should have one Addr and the rest should be WeakAddrs

Sandipmandal25:
> > the watcher closure also holds a strong addr so when we set the parent field to none the actor still has one ref alive in the watcher. should we remove the watcher too or is there something that handles this.?
> 
> We should have one Addr and the rest should be WeakAddrs

fixed!

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
```diff
@@ -423,6 +423,30 @@ private fun SendFlowRouteToScreen(
                 }
             }
 
+            // show success UI on payjoin broadcast; TaggedItem key changes each time so
+            // no manual reset is needed even if the user sends multiple payjoin transactions
+            LaunchedEffect(walletManager.payjoinTxBroadcast) {
+                if (walletManager.payjoinTxBroadcast != null && sendState == SendState.Sending) {
+                    sendState = SendState.Sent
+                    showSuccessAlert = true
+                    Auth.unlock()
```
praveenperera, line 468: why auth.unlock?

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,341 @@
...
+                    send!(wallet_addr.handle_payjoin_success(proposal_psbt, fallback_tx));
+                }
+                Ok(OptionalTransitionOutcome::Stasis(next)) => {
+                    debug!("payjoin poll: no proposal yet, continuing");
+                    send!(addr.update_polling_sender(next));
+                }
+                Err(e) => {
+                    warn!("payjoin poll: fatal error processing response: {e:?}");
+                    send!(wallet_addr.handle_payjoin_fallback(fallback_tx));
+                }
```
praveenperera: This treats every `process_response(...).save(...)` error as fatal and broadcasts the fallback transaction, but the Payjoin transition type distinguishes transient directory/OHTTP failures from fatal session failures. 

A single transient bad relay response can cancel an otherwise valid Payjoin attempt and leak back to a normal spend. 

Match the persisted transition error so transient errors log and schedule `begin_next_poll_msg()` using the still-stored polling sender, and only fatal errors close the session and broadcast fallback.

ios/Cove/Flows/SendFlow/SendFlowConfirmScreen.swift
```diff
@@ -184,6 +184,13 @@ struct SendFlowConfirmScreen: View {
                                     psbt: details.psbt(),
                                     payjoinEndpoint: payjoinEndpoint
                                 )
+                                // for payjoin, stay in .sending — PayjoinTxBroadcast reconcile fires sendState = .sent
+                                if payjoinEndpoint == nil {
+                                    sendState = .sent
+                                    isShowingAlert = true
+                                    auth.unlock()
```
praveenperera, line 191: here too?

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,341 @@
...
+            };
+
+            let (post_response, post_ctx) = match post_result {
+                Ok(pair) => pair,
+                Err(e) => {
+                    warn!("payjoin POST: all relays failed: {e}");
+                    send!(wallet_addr.handle_payjoin_fallback(fallback_tx));
+                    return;
+                }
+            };
```
praveenperera: too much nesting

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,341 @@
...
+                        }
+                        Err(e) => {
+                            warn!("payjoin: relay {relay} poll failed: {e}");
+                            last_err = e;
+                        }
+                    }
+                }
+                success.ok_or(last_err)
+            };
+
```
praveenperera: too much nesting

praveenperera (changes requested):

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
Sandipmandal25: same as normal send we lock at start and unlock when done payjoin just takes longer to complete so unlock happens via the reconcile instead.

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: yeah not all errors are fatal if receiver explicitly rejects we fallback otherwise retry since we still have the polling sender stored.

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: extracted the relay loop into a shared helper function both post and poll use it now

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,343 @@
...
+    let bip21 = format!("bitcoin:{address}?amount={amount_btc:.8}&pj={encoded_endpoint}");
+    let pj_uri = Uri::try_from(bip21.as_str())
+        .map_err(|e| eyre::eyre!("failed to parse payjoin URI: {e:?}"))?
+        .assume_checked()
+        .check_pj_supported()
+        .map_err(|_| eyre::eyre!("URI does not support payjoin (missing pj= param)"))?;
+
+    let persister = NoopSessionPersister::<PayjoinSessionEvent>::default();
+    let sender = SenderBuilder::new(signed_psbt, pj_uri)
```
praveenperera, line 161: When a scanned BIP21 contains a v1 `pj=` endpoint, `AddressWithNetwork` still records it and Cargo now enables both payjoin `v1` and `v2`, so `check_pj_supported()` can return a V1 URI here. Passing that into `payjoin::send::v2::SenderBuilder::new` hits the crate's `unimplemented!` V1 arm, panicking instead of falling back to a normal/fallback broadcast. Reject non-v2 endpoints before this call or avoid routing v1 endpoints into the v2 sender.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,343 @@
...
+                        return;
+                    }
+                };
+
+            let persister = NoopSessionPersister::<PayjoinSessionEvent>::default();
+            match polling_sender.process_response(&poll_response, poll_ctx).save(&persister) {
+                Ok(OptionalTransitionOutcome::Progress(proposal_psbt)) => {
+                    send!(wallet_addr.handle_payjoin_success(proposal_psbt, fallback_tx));
+                }
+                Ok(OptionalTransitionOutcome::Stasis(next)) => {
```
praveenperera: When the receiver never publishes a proposal, the directory can keep returning the no-result/Stasis response, and this branch immediately continues polling forever with no elapsed-time or URI-expiration limit. In that scenario the UI remains in Sending and the original transaction is never broadcast; add an overall timeout/expiration path that calls `handle_payjoin_fallback`.

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -727,6 +772,64 @@ impl WalletActor {
...
+                }
+            },
+            Err(error) => {
+                error!("failed to sign payjoin proposal, falling back to original tx: {error:?}");
+                self.do_broadcast_transaction(fallback_tx).await
+            }
+        };
+
+        match broadcast_result {
+            Ok(()) => self.send(WalletManagerReconcileMessage::PayjoinTxBroadcast),
```
praveenperera: When a payjoin proposal or fallback broadcast succeeds, this success event is reached after calling only `do_broadcast_transaction`; unlike the direct send/broadcast paths, the actual transaction is never passed to `insert_broadcast_transaction`. The UI can show success while the balance, transaction list, and confirmation watcher remain stale until a later scan, so insert the broadcast proposal/fallback tx before emitting `PayjoinTxBroadcast`.

praveenperera (changes requested):

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: fixed reject v1 endpoints before reaching the v2 sender builder

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: added a session deadline that triggers fallback broadcast when exceeded

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: inserted broadcast transaction before emitting the reconcile in both handlers

Sandipmandal25:
@coderabbitai  re review

Sandipmandal25:
@greptileai re review

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,482 @@
...
+                warn!("payjoin: relay {relay} request failed: {e:?}");
+                last_err = e;
+            }
+        }
+    }
+    Err(last_err)
+}
+
+/// Opaque alias so `actor.rs` can name the initial sender state without importing payjoin internals
+pub(super) type PayjoinSender = V2Sender<WithReplyKey>;
```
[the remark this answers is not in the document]
Sandipmandal25: pub(super) scopes it to the parent module only which is what we want here. pub(crate) would be too wide thats what i think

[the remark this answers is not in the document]
Sandipmandal25: fixed!

Sandipmandal25:
@coderabbitai re review

ios/Cove/Flows/SendFlow/SendFlowConfirmScreen.swift
```diff
@@ -217,6 +224,24 @@ struct SendFlowConfirmScreen: View {
...
+                // payjoin broadcast failure arrives via reconcile (not the catch block),
+                // so we must handle it here to unblock the UI from .sending
+                guard let alert, case .sending = sendState else { return }
+                let errorMessage =
+                    switch alert.item {
+                    case let .signAndBroadcast(error): error
+                    case let .confirmDetails(error): error
+                    }
+                sendState = .error(errorMessage)
+            }
```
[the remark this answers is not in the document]
Sandipmandal25: the container already shows the error alert for the payjoin path. setting it here too was the duplicate alert

Sandipmandal25:
@greptileai

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -785,6 +836,80 @@ impl WalletActor {
...
+                }
+            }
+            Err(error) => {
+                error!("failed to sign payjoin proposal, falling back to original tx: {error:?}");
+                match self.do_broadcast_transaction(fallback_tx.clone()).await {
+                    Ok(()) => (Ok(()), Some(fallback_tx)),
+                    Err(e) => (Err(e), None),
+                }
+            }
+        };
```
praveenperera: why is this in a block?

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -635,18 +638,57 @@ impl WalletActor {
...
+        let network: bitcoin::Network = self.wallet.network.into();
+
+        let sender = match build_sender(signed_psbt, &fallback_tx, endpoint, network) {
+            Ok(s) => s,
+            Err(e) => {
+                warn!("payjoin setup failed, broadcasting fallback tx: {e}");
+                send!(self.addr.handle_payjoin_fallback(fallback_tx));
+                return Ok(());
+            }
+        };
```
praveenperera: you used a let else earlier but a match here, any reason why? its fine just curious, we should probably stay consistent when it makes sense tho.

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -635,18 +638,57 @@ impl WalletActor {
...
+            )));
+        }
+
+        let Some(endpoint) = payjoin_endpoint else {
+            // non-payjoin path: synchronous sign and broadcast
+            let result = self.do_sign_and_broadcast_transaction(psbt).await;
+            return Produces::ok(result);
+        };
+
+        let result = self.initiate_payjoin_payment(psbt, endpoint).await;
```
praveenperera: i think a match here actually makes more sense because the the 2 branches are 2 different things, one branch is payjoin other is normal txn and we can add a comment

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -676,11 +718,20 @@ impl WalletActor {
             return Err(err("transaction not finalized, unable to sign"));
         }
 
-        let transaction = psbt
-            .extract_tx()
+        // clone before extract so we retain the signed PSBT for the payjoin POST;
+        // extract_tx_fee_rate_limit rejects extraction if the fee rate is unreasonably high
```
praveenperera: why this note? about it rejecting extraction? whats the significance also waht does it count as unreasonably high?

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,569 @@
...
+        .to_vec())
+}
+
+// returns OHTTP relay URLs shuffled per call for resilience and privacy
+fn ohttp_relays() -> Vec<&'static str> {
+    let mut relays = vec![
+        "https://relay.payjoin.org",
+        "https://ohttp.achow101.com",
+        "https://pj.bobspacebkk.com",
+    ];
```
praveenperera: lets put this into a const at the top of the file or

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,569 @@
...
+    relays.shuffle(&mut rand::rng());
+    relays
+}
+
+// tries each OHTTP relay in shuffled order, returning the first successful (body, context) pair
+async fn try_ohttp_relays<C>(
+    client: &reqwest::Client,
+    timeout: Duration,
+    build_request: impl Fn(&str) -> eyre::Result<(payjoin::Request, C)>,
+) -> eyre::Result<(Vec<u8>, C)> {
```
praveenperera: we usually just do use eyre::Result at the top of the file

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,569 @@
...
+                warn!("payjoin: relay {relay} request failed: {e:?}");
+                last_err = e;
+            }
+        }
+    }
+    Err(last_err)
+}
+
+/// Opaque alias so `actor.rs` can name the initial sender state without importing payjoin internals
+pub(crate) type PayjoinSender = V2Sender<WithReplyKey>;
```
praveenperera: put with the types at the top of the file?

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,569 @@
...
+        let polling_sender = match &self.session {
+            PayjoinSession::Polling { polling_sender } => polling_sender.clone(),
+            _ => {
+                warn!("payjoin begin_poll called in unexpected state");
+                return Produces::ok(());
+            }
+        };
+
+        self.addr.send_fut_with(|addr| async move {
+            let client = match cove_http::new_client() {
```
praveenperera: lets extract the inside into a function

rust/src/manager/wallet_manager/payjoin.rs
praveenperera, line 1: does it make sense to split the actors stuff into `wallet_manager/payjoin/actor.rs` i'm not sure.

praveenperera (changes requested):

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,569 @@
...
+        .timeout(timeout)
+        .send()
+        .await
+        .map_err(|e| eyre::eyre!("send failed: {e:?}"))?
+        .error_for_status()
+        .map_err(|e| eyre::eyre!("relay returned error status: {e:?}"))?
+        .bytes()
+        .await
+        .map_err(|e| eyre::eyre!("body read failed: {e:?}"))?
+        .to_vec())
```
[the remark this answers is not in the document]
Sandipmandal25: the relay response is always a small psbt we re only hitting trusted payjoin infra and theres already a timeout bounding the read adding streaming complexity here isnt worth it

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -0,0 +1,569 @@
...
+
+    let idx = recipient_output_index(&signed_psbt)?;
+
+    let txout = &signed_psbt.unsigned_tx.output[idx];
+    let amount_btc = Amount::from_sat(txout.value.to_sat()).to_btc();
+    let address = bitcoin::Address::from_script(&txout.script_pubkey, Params::from(network))
+        .map_err(|e| eyre::eyre!("could not derive address from PSBT output: {e:?}"))?;
+
+    let bip21 = format!("bitcoin:{address}?amount={amount_btc:.8}&pj={encoded_endpoint}");
+    let pj_uri = Uri::try_from(bip21.as_str())
```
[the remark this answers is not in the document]
Sandipmandal25: bip77 v2 ohttp endpoints dont have & or = in the url path only # which we already handle. form encoding would over encode and break the uri parsing downstream

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: bdk hardcodes it at 25000 sat/vbyte basically a sanity check against accidental absurd fees. anyway dropping that note!

Sandipmandal25:
@greptileai

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: the two are logically distinct so the split makes sense but worried about breaking things can we do it as a follow up? or should i do it in this one?

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: fixed

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: fixed

rust/src/manager/wallet_manager/payjoin.rs
Sandipmandal25: fixed!

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: changed to match with comments on each branch

rust/src/manager/wallet_manager/actor.rs
Sandipmandal25: extracted a helper and rewrote this without the block

rust/src/manager/wallet_manager/payjoin.rs
praveenperera: we can do it later

praveenperera:
merged thanks @sandipmandal25

PR #809, opened 2026-06-25
Persist payjoin sender session and resume on cold restart

Sandipmandal25:
## summary

this pr makes payjoin sender sessions recoverable after a cold restart.

the payjoin sender pr (#772) used `NoopSessionPersister`, so any in flight payjoin session was lost if the app restarted before the negotiation finished. this implements a real `SessionPersister` backed by `redb`.

## changes

the session event log is now persisted before the first network request, along with the consensus-encoded fallback transaction. on startup, the stored event log is replayed so the app can continue the payjoin flow safely.

depending on the recovered state, the sender can now resume polling, broadcast an already completed proposal, or fall back to the original transaction if the session cannot be reconstructed.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Payjoin sessions now persist across app restarts, letting in-progress send flows resume automatically.
  * Added transaction lookup support (mempool-or-chain) to improve terminal handling during send recovery.

* **Bug Fixes**
  * Improved recovery for saved payjoin sessions, including more reliable fallback/proposal resumption and protection against corrupt saved state.
  * More consistent session cleanup and safer retry behavior when persistence or broadcasting fails.

### DIFF
--- a/rust/src/database/wallet_data.rs
+++ b/rust/src/database/wallet_data.rs
@@ -86,15 +86,17 @@ impl ReceiveAddressCache {
 /// `bitcoin::Transaction` does not implement serde natively; storing the consensus-encoded
 /// bytes avoids an orphan-impl problem while keeping the format identical to what is broadcast.
 #[derive(
-    Debug, Clone, Serialize, Deserialize, Eq, PartialEq, derive_more::From, derive_more::Into,
+    Debug,
+    Clone,
+    Serialize,
+    Deserialize,
+    Eq,
+    PartialEq,
+    derive_more::From,
+    derive_more::Into,
+    derive_more::AsRef,
 )]
 pub struct TransactionBytes(Vec<u8>);
-
-impl AsRef<[u8]> for TransactionBytes {
-    fn as_ref(&self) -> &[u8] {
-        &self.0
-    }
-}
 
 /// Terminal action decided for a payjoin session before the broadcast completes
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
--- a/rust/src/database/wallet_data.rs
+++ b/rust/src/database/wallet_data.rs
@@ -81,18 +81,33 @@ impl ReceiveAddressCache {
     }
 }
 
+/// Consensus-encoded bytes of a bitcoin transaction stored in the database.
+///
+/// `bitcoin::Transaction` does not implement serde natively; storing the consensus-encoded
+/// bytes avoids an orphan-impl problem while keeping the format identical to what is broadcast.
+#[derive(
+    Debug, Clone, Serialize, Deserialize, Eq, PartialEq, derive_more::From, derive_more::Into,
+)]
+pub struct TransactionBytes(Vec<u8>);
+
+impl AsRef<[u8]> for TransactionBytes {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
 /// Terminal action decided for a payjoin session before the broadcast completes
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub enum PendingAction {
     BroadcastFallback,
-    BroadcastProposal { transaction: Vec<u8> },
+    BroadcastProposal { transaction: TransactionBytes },
 }
 
 /// Event log for an in-flight payjoin sender session, allowing it to survive app restarts
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub struct PayjoinSenderSession {
     pub events: Vec<String>,
-    pub fallback_tx: Vec<u8>,
+    pub fallback_tx: TransactionBytes,
     #[serde(default)]
     pub created_at_secs: Option<u64>,
     #[serde(default)]
@@ -442,7 +457,7 @@ mod tests {
         let (db, _tmp) = test_support::new_test_wallet_data_db(wallet_id);
         let session = PayjoinSenderSession {
             events: vec![r#"{"PostedOriginalPsbt":[]}"#.to_string()],
-            fallback_tx: vec![0x02, 0x00, 0x00, 0x00],
+            fallback_tx: vec![0x02, 0x00, 0x00, 0x00].into(),
             created_at_secs: None,
             pending_action: None,
         };
@@ -458,7 +473,7 @@ mod tests {
         let (db, _tmp) = test_support::new_test_wallet_data_db(wallet_id);
         let session = PayjoinSenderSession {
             events: vec![r#"{"Closed":"Failure"}"#.to_string()],
-            fallback_tx: vec![0x02, 0x00, 0x00, 0x00],
+            fallback_tx: vec![0x02, 0x00, 0x00, 0x00].into(),
             created_at_secs: None,
             pending_action: None,
         };
--- a/rust/src/database/wallet_data.rs
+++ b/rust/src/database/wallet_data.rs
@@ -85,6 +85,7 @@ impl ReceiveAddressCache {
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub enum PendingAction {
     BroadcastFallback,
+    BroadcastProposal { transaction: Vec<u8> },
 }
 
 /// Event log for an in-flight payjoin sender session, allowing it to survive app restarts
--- a/rust/src/database/wallet_data.rs
+++ b/rust/src/database/wallet_data.rs
@@ -42,12 +42,14 @@ pub enum WalletData {
     /// number of addresses scanned
     ScanState(ScanState),
     ReceiveAddressCache(ReceiveAddressCache),
+    PayjoinSenderSession(PayjoinSenderSession),
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, uniffi::Enum)]
 pub enum WalletDataKey {
     ScanState(WalletAddressType),
     ReceiveAddressCache,
+    PayjoinSenderSession,
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, uniffi::Enum)]
@@ -77,6 +79,23 @@ impl ReceiveAddressCache {
         self.first_shown_at_secs = now_secs;
         self
     }
+}
+
+/// Terminal action decided for a payjoin session before the broadcast completes
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub enum PendingAction {
+    BroadcastFallback,
+}
+
+/// Event log for an in-flight payjoin sender session, allowing it to survive app restarts
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct PayjoinSenderSession {
+    pub events: Vec<String>,
+    pub fallback_tx: Vec<u8>,
+    #[serde(default)]
+    pub created_at_secs: Option<u64>,
+    #[serde(default)]
+    pub pending_action: Option<PendingAction>,
 }
 
 #[derive(Debug, Clone, uniffi::Object)]
@@ -177,6 +196,24 @@ impl WalletDataDb {
 
     pub fn delete_receive_address_cache(&self) -> Result<()> {
         self.delete(WalletDataKey::ReceiveAddressCache)
+    }
+
+    pub fn get_payjoin_sender_session(&self) -> Result<Option<PayjoinSenderSession>> {
+        let value = self.get(WalletDataKey::PayjoinSenderSession)?;
+
+        let Some(WalletData::PayjoinSenderSession(session)) = value else {
+            return Ok(None);
+        };
+
+        Ok(Some(session))
+    }
+
+    pub fn set_payjoin_sender_session(&self, session: PayjoinSenderSession) -> Result<()> {
+        self.set(WalletDataKey::PayjoinSenderSession, WalletData::PayjoinSenderSession(session))
+    }
+
+    pub fn delete_payjoin_sender_session(&self) -> Result<()> {
+        self.delete(WalletDataKey::PayjoinSenderSession)
     }
 
     fn get(&self, key: WalletDataKey) -> Result<Option<WalletData>> {
@@ -298,6 +335,7 @@ impl WalletDataKey {
             Self::ScanState(WalletAddressType::WrappedSegwit) => "scan_state_wrapped_segwit",
             Self::ScanState(WalletAddressType::Legacy) => "scan_state_legacy",
             Self::ReceiveAddressCache => "receive_address_cache",
+            Self::PayjoinSenderSession => "payjoin_sender_session",
         }
     }
 }
@@ -395,5 +433,48 @@ mod tests {
         db.delete_receive_address_cache().unwrap();
 
         assert_eq!(db.get_receive_address_cache().unwrap(), None);
+    }
+
+    #[test]
+    fn payjoin_sender_session_round_trips() {
+        let wallet_id = WalletId::preview_new_random();
+        let (db, _tmp) = test_support::new_test_wallet_data_db(wallet_id);
+        let session = PayjoinSenderSession {
+            events: vec![r#"{"PostedOriginalPsbt":[]}"#.to_string()],
+            fallback_tx: vec![0x02, 0x00, 0x00, 0x00],
+            created_at_secs: None,
+            pending_action: None,
+        };
+
+        db.set_payjoin_sender_session(session.clone()).unwrap();
+
+        assert_eq!(db.get_payjoin_sender_session().unwrap(), Some(session));
+    }
+
+    #[test]
+    fn delete_payjoin_sender_session_clears_session() {
+        let wallet_id = WalletId::preview_new_random();
+        let (db, _tmp) = test_support::new_test_wallet_data_db(wallet_id);
+        let session = PayjoinSenderSession {
+            events: vec![r#"{"Closed":"Failure"}"#.to_string()],
+            fallback_tx: vec![0x02, 0x00, 0x00, 0x00],
+            created_at_secs: None,
+            pending_action: None,
+        };
+
+        db.set_payjoin_sender_session(session).unwrap();
+        db.delete_payjoin_sender_session().unwrap();
+
+        assert_eq!(db.get_payjoin_sender_session().unwrap(), None);
+    }
+
+    #[test]
+    fn delete_missing_payjoin_sender_session_succeeds() {
+        let wallet_id = WalletId::preview_new_random();
+        let (db, _tmp) = test_support::new_test_wallet_data_db(wallet_id);
+
+        db.delete_payjoin_sender_session().unwrap();
+
+        assert_eq!(db.get_payjoin_sender_session().unwrap(), None);
     }
 }
--- a/rust/src/manager/wallet_manager/actor.rs
+++ b/rust/src/manager/wallet_manager/actor.rs
@@ -211,6 +211,13 @@ impl WalletActor {
         Produces::ok(())
     }
 
+    pub async fn notify_payjoin_error(&mut self, msg: String) -> ActorResult<()> {
+        self.send(WalletManagerReconcileMessage::SendFlowError(
+            SendFlowErrorAlert::SignAndBroadcast(msg),
+        ));
+        Produces::ok(())
+    }
+
     #[into_actor_result]
     pub async fn unlocked_trusted_spendable_balance(&mut self) -> Result<Amount, Error> {
         self.unlocked_trusted_spendable_balance_inner()
--- a/rust/src/manager/wallet_manager/actor.rs
+++ b/rust/src/manager/wallet_manager/actor.rs
@@ -17,7 +17,7 @@ mod scan;
 mod transaction_confirmation;
 mod transactions;
 
-use super::payjoin::PayjoinActor;
+use super::payjoin::{PayjoinActor, SessionResumption, resume_session};
 use act_zero::{runtimes::tokio::spawn_actor, *};
 use act_zero_ext::into_actor_result;
 use ahash::HashMap;
@@ -108,6 +108,7 @@ impl Actor for WalletActor {
         self.addr = addr.downgrade();
         self.spawn_scan_actor();
         send!(addr.check_node_connection());
+        send!(addr.resume_payjoin_session());
         Produces::ok(())
     }
 
@@ -183,6 +184,31 @@ impl WalletActor {
     pub async fn balance(&mut self) -> ActorResult<Balance> {
         let balance = self.wallet.balance();
         Produces::ok(balance)
+    }
+
+    /// Resumes a persisted payjoin session from a previous app run, if one exists
+    pub async fn resume_payjoin_session(&mut self) -> ActorResult<()> {
+        if self.payjoin_actor.is_some() {
+            return Produces::ok(());
+        }
+
+        match resume_session(self.db.clone(), self.addr.clone()) {
+            SessionResumption::None => {}
+
+            SessionResumption::Resume(actor) => {
+                self.payjoin_actor = Some(spawn_actor(*actor));
+            }
+
+            SessionResumption::BroadcastProposal { proposal_psbt, fallback_tx } => {
+                send!(self.addr.handle_payjoin_success(proposal_psbt, fallback_tx));
+            }
+
+            SessionResumption::BroadcastFallback { fallback_tx } => {
+                send!(self.addr.handle_payjoin_fallback(fallback_tx));
+            }
+        }
+
+        Produces::ok(())
     }
 
     #[into_actor_result]
--- a/rust/src/manager/wallet_manager/actor.rs
+++ b/rust/src/manager/wallet_manager/actor.rs
@@ -199,12 +199,20 @@ impl WalletActor {
                 self.payjoin_actor = Some(spawn_actor(*actor));
             }
 
+            SessionResumption::BroadcastStoredProposal { proposal_tx } => {
+                send!(self.addr.handle_payjoin_proposal_broadcast(proposal_tx));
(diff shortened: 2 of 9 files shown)


Sandipmandal25:
@coderabbitai

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -650,6 +659,10 @@ impl WalletActor {
 
         match self.broadcast_transaction_to_node(proposal_tx.clone()).await {
             Ok(()) => {
+                if let Err(error) = self.db.delete_payjoin_sender_session() {
+                    warn!("failed to clear payjoin session record: {error}");
```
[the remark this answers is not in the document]
Sandipmandal25: Fixed session is now cleared on every successful terminal broadcast.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -247,24 +316,47 @@ pub(crate) struct PayjoinActor {
...
+        polling_sender: V2Sender<PollingForProposal>,
+        fallback_tx: BdkTransaction,
+    ) -> Self {
+        Self {
+            addr: WeakAddr::default(),
+            wallet_addr,
+            persister,
+            fallback_tx,
+            session: PayjoinSession::Polling { polling_sender },
+            poll_deadline: Some(Instant::now() + PAYJOIN_SESSION_TIMEOUT),
```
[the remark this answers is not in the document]
Sandipmandal25: creation time is persisted and remaining deadline is computed on resume. Expired sessions fall back immediately

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -650,6 +659,10 @@ impl WalletActor {
...
+                }
+
```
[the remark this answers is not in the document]
Sandipmandal25: Cleaned up the fallback paths. Skipped the "already broadcast" case  we cant reliably detect it from the nodes error response

Sandipmandal25:
@greptileai

Sandipmandal25:
@greptileai

Sandipmandal25:
@greptileai

Sandipmandal25:
@greptileai

Sandipmandal25:
@coderabbitai re review

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -106,6 +106,7 @@ impl Actor for WalletActor {
         self.addr = addr.downgrade();
         self.spawn_scan_actor();
         send!(addr.check_node_connection());
+        send!(addr.resume_payjoin_session());
```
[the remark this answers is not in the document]
Sandipmandal25: Partially addressed send gate now attempts local cleanup before blocking (terminal marker and tx in wallet to delete inline no restart needed). Full offline node retry loop deferred as followup session safely is retained!

Sandipmandal25:
@greptileai

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -455,13 +693,158 @@ impl PayjoinActor {
...
+        Ok(fallback_tx) => SessionResumption::BroadcastFallback { fallback_tx },
+        Err(error) => {
+            error!("stored payjoin fallback tx is corrupt, abandoning session: {error}");
+            if let Err(error) = db.delete_payjoin_sender_session() {
+                warn!("failed to clear corrupt payjoin session record: {error}");
+            }
+            SessionResumption::None
+        }
+    }
+}
```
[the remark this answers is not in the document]
Sandipmandal25: but i think is this is slightly too broad if there is no terminal marker the fallback was never dispatched so retaining that session would permanently block future sends with no way to recover ig better fix is it to be context sensitive based on whether the marker was written. done that!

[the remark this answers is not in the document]
Sandipmandal25: thats a good way to put it no need to save as a rule!

Sandipmandal25:
@greptileai

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -180,6 +181,46 @@ impl WalletActor {
...
+            SessionResumption::ReportError { message } => {
+                send!(self.addr.notify_payjoin_error(message));
+            }
+        }
+
+        Produces::ok(())
+    }
+
+    pub async fn notify_payjoin_error(&mut self, msg: String) -> ActorResult<()> {
+        self.send(WalletManagerReconcileMessage::SendFlowError(
```
praveenperera: Can we surface startup recovery failures through the app-level alert path instead of `SendFlowError`? `resume_payjoin_session()` runs from `started()`, when the send flow usually is not mounted. On iOS this is only presented by `SendFlowContainer`, and on Android the send flow only turns it into visible UI while already in `Sending`, so a failed recovery/rebroadcast can leave the session retained with no visible explanation. The next normal send then just hits the pending-recovery gate.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -26,8 +34,183 @@ const PAYJOIN_SESSION_TIMEOUT: Duration = Duration::from_secs(10 * 60);
 const OHTTP_RELAYS: [&str; 3] =
     ["https://relay.payjoin.org", "https://ohttp.achow101.com", "https://pj.bobspacebkk.com"];
 
-/// Opaque alias so `actor.rs` can name the initial sender state without importing payjoin internals
-pub(crate) type PayjoinSender = V2Sender<WithReplyKey>;
+/// Persists payjoin session events to the wallet's database so sessions survive app restarts.
+///
+/// `lock` is shared across all clones and serialises `save_event` against `set_pending_fallback`
```
praveenperera: This comment overstates what the lock protects. `PayjoinSessionPersister::new()` allocates a fresh `Arc<Mutex<()>>`, so independently constructed persisters do not share a lock; only clones of the same instance do. I think the safety here currently comes from the pending-action marker/check and actor ordering, not from process-wide serialization. Either the lock should be shared structurally, or this comment should describe the narrower invariant.

Also when I see `Mutexes` I ask myself, would an actor be better solution to this problem.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -26,8 +34,183 @@ const PAYJOIN_SESSION_TIMEOUT: Duration = Duration::from_secs(10 * 60);
...
+            .map_err(|_| WalletDataError::Save("session lock poisoned".to_string()))?;
+
+        let mut session = self
+            .db
+            .get_payjoin_sender_session()?
+            .ok_or_else(|| WalletDataError::Save("no payjoin session record".to_string()))?;
+
+        // Reject late events once any terminal intent is committed: an in-flight poll
+        // response must not overwrite a BroadcastFallback or BroadcastProposal marker.
+        // The shared lock ensures this check-then-write is atomic with set_pending_action.
```
praveenperera: Same concern here: this lock is only shared with clones of this persister instance, not with later `PayjoinSessionPersister::new(self.db.clone())` instances in the wallet actor. Also, the comment references `set_pending_action`, but the actual methods are `set_pending_fallback` and `set_pending_proposal`.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -455,13 +693,174 @@ impl PayjoinActor {
...
+        Ok(pair) => pair,
+        Err(error) => {
+            warn!("payjoin session replay failed, broadcasting fallback: {error:?}");
+            return fallback_from_record(&db, record);
+        }
+    };
+
+    let fallback_tx = history.fallback_tx();
+    match session {
+        SendSession::WithReplyKey(sender) => SessionResumption::Resume(Box::new(
```
praveenperera: This recovered state is ambiguous. If the app crashes after the v2 POST reaches the directory but before `PostedOriginalPsbt` is saved, replay lands back in `WithReplyKey`; spawning `PayjoinActor::new` then calls `start_post()` and sends the original PSBT again. Since sender POSTs are not retry-safe, I think this should fall back here, or persist a pre-POST attempt marker before the network request.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -455,13 +693,174 @@ impl PayjoinActor {
...
+                wallet_addr,
+                persister,
+                polling_sender,
+                fallback_tx,
+                record.created_at_secs,
+            )))
+        }
+
+        SendSession::Closed(SessionOutcome::Success(proposal_psbt)) => {
+            SessionResumption::SignRecoveredProposal { proposal_psbt }
```
praveenperera: Could we keep the signed fallback on this recovery path? If the app crashes after `Closed(Success)` is saved but before `set_pending_proposal()` stores the signed proposal tx, a later signing failure for the recovered proposal only retains the session. The normal in-memory path can still broadcast the already-signed fallback before any proposal tx is committed; after restart this path has dropped it, so a bad/unsignable recovered proposal can block future sends unnecessarily.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -455,13 +693,174 @@ impl PayjoinActor {
...
+    Resume(Box<PayjoinActor>),
+    /// A BroadcastProposal marker was stored: broadcast this exact consensus-encoded tx
+    BroadcastStoredProposal { proposal_tx: BdkTransaction },
+    /// Session closed with a success outcome but no stored tx: sign the recovered PSBT and broadcast.
+    /// `fallback_tx` is intentionally omitted — a local signing failure during recovery must pause
+    /// and preserve the record rather than downgrade to the original transaction.
+    SignRecoveredProposal { proposal_psbt: Psbt },
+    /// Session ended without a proposal, broadcast the original tx
+    BroadcastFallback { fallback_tx: BdkTransaction },
+    /// The stored proposal tx is corrupt; the record is retained and the user must be notified.
```
praveenperera: This variant is broader than the doc says. `ReportError` is also returned for an uncommitted corrupt fallback, and that path deletes the record because no terminal marker was written. I would make the enum-level contract generic and leave the "record is retained" guarantee on the specific proposal/committed-fallback producers.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -455,13 +693,174 @@ impl PayjoinActor {
...
+/// Recovers the proposal tx committed in a `BroadcastProposal` marker so it can be
+/// re-broadcast on the next startup without re-signing.
+///
+/// If the stored bytes are corrupt the record is **retained** and `ReportError` is returned.
+/// We cannot know whether that proposal reached the network before the corruption/restart,
+/// so deleting the record or broadcasting the fallback would be unsafe.
+fn proposal_from_record(_db: &WalletDataDb, record: PayjoinSenderSession) -> SessionResumption {
+    let tx_bytes = match &record.pending_action {
+        Some(PendingAction::BroadcastProposal { transaction }) => transaction.clone(),
+        _ => unreachable!("proposal_from_record called without BroadcastProposal action"),
```
praveenperera: I would avoid a panic in this recovery path. It is unreachable with the current callers, but this whole flow is deliberately fail-closed; returning `ReportError` here keeps a future refactor mistake from turning recovery into a crash.

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -439,6 +439,53 @@ impl WalletActor {
             )));
         }
 
+        match self.db.get_payjoin_sender_session() {
```
praveenperera, line 457: Can we add actor-level coverage around this send gate? This is what blocks every future send while recovery state exists, and it also has the cleanup path for the "terminal tx already persisted locally" case. The lower-level persister/resume matrix is covered, but I do not see a test proving this pre-check false-rejects/false-allows correctly.

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -618,17 +674,144 @@ impl WalletActor {
...
+        self.send(Msg::UpdatedTransactions(transactions));
+
+        send!(self.addr.start_transaction_watcher(txid));
+
+        if session_cleared {
+            self.send(Msg::PayjoinTxBroadcast);
+        } else {
+            // TX is broadcast and wallet-persisted, but the session record remains.
+            // initiate_payment will reject new sends until the record is gone.
+            // The next startup will clear it automatically via the wallet pre-check.
```
praveenperera: This comment points at the wrong mechanism. Startup cleanup happens through `resume_payjoin_session()` and then the terminal broadcast/finalize path; the wallet pre-check only runs when the user starts another send, and in that case it tries to clean up and allow the send rather than just asking for a restart.

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -618,17 +674,144 @@ impl WalletActor {
...
+    /// queries the node: if the tx is already in the mempool or chain (i.e., a prior broadcast
+    /// succeeded but the app crashed before the session record was cleared), finalize normally
+    /// rather than leaving a stuck session that blocks future sends.  If the tx is genuinely
+    /// absent or the query itself fails, retain the session record for the next-startup retry.
+    async fn broadcast_payjoin_terminal(&mut self, tx: BdkTransaction) {
+        let txid = tx.compute_txid();
+
+        // Wallet already has the tx — a prior run persisted it but crashed before clearing the
+        // session record.  Skip rebroadcast and go straight to cleanup.
+        if self.wallet.bdk.get_tx(txid).is_some() {
```
praveenperera: This branch seems worth an actor-level test too. It is the important "tx already in the wallet, skip rebroadcast, then finalize/clear" path and does not need a live node to exercise.

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -643,24 +826,87 @@ impl WalletActor {
...
+        proposal_psbt: Psbt,
+    ) -> ActorResult<()> {
+        let proposal_tx = match self.do_sign_original_psbt(proposal_psbt).await {
+            Ok((_, tx)) => tx,
             Err(error) => {
-                error!(
-                    "failed to broadcast payjoin proposal, falling back to original tx: {error:?}"
-                );
-                self.broadcast_and_notify(fallback_tx).await;
+                error!("failed to sign recovered payjoin proposal, pausing for retry: {error:?}");
```
praveenperera, line 884: Please add a focused test for this failure path. This is one of the highest-risk invariants in the recovery flow: once a receiver proposal has been accepted, a recovered signing failure must retain the record and must not fall back to the original transaction.

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -26,8 +34,183 @@ const PAYJOIN_SESSION_TIMEOUT: Duration = Duration::from_secs(10 * 60);
...
+            .ok_or_else(|| WalletDataError::Save("no payjoin session record".to_string()))?;
+
+        match &session.pending_action {
+            None => {}
+
+            Some(PendingAction::BroadcastProposal { transaction }) if *transaction == tx_bytes => {
+                return Ok(());
+            }
+
+            Some(_) => {
```
praveenperera, line 110: The idempotency test covers re-saving the same proposal tx, but I think we should also test the anti-overwrite case where a different proposal tx is rejected. That is the guard that prevents replacing a committed terminal action with different bytes.

rust/src/node/client/electrum.rs
```diff
@@ -166,6 +166,21 @@ impl ElectrumClient {
...
     }
 
+    /// Fetches a transaction from the mempool or chain; returns `None` if not found.
+    pub async fn get_transaction(&self, txid: Txid) -> Result<Option<bitcoin::Transaction>, Error> {
+        let client = self.client.clone();
+        cove_tokio::unblock::run_blocking(move || client.inner.transaction_get(&txid))
+            .await
+            .map(Some)
+            .or_else(|e| match e {
+                // Only a null response is a definitive "not found"; all other
```
praveenperera: This does not match the common Electrum not-found shape. `electrum-client` documents `transaction_get` as returning an error if the tx is not found, and JSON-RPC error objects come through as `Error::Protocol`, not `InvalidResponse(Value::Null)`. The current recovery caller treats `Err` fail-safe, so this is not unsafe today, but `get_transaction` now has different `Ok(None)` semantics from Esplora and can mislead a future caller that distinguishes "absent" from "couldn't tell".

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -618,17 +674,144 @@ impl WalletActor {
...
+    /// wallet, persist, clear the session record, and update the UI.  The session record is
+    /// deleted only after the wallet is durably persisted so a crash between broadcast and
+    /// persist still has a recovery path on the next startup.
+    async fn finalize_payjoin_terminal(&mut self, tx: BdkTransaction) {
+        use WalletManagerReconcileMessage as Msg;
+
+        let now =
+            SystemTime::now().duration_since(UNIX_EPOCH).map(|d| d.as_secs()).unwrap_or_else(|e| {
+                warn!("System clock skew detected: {e}");
+                u64::MAX
```
praveenperera: Minor, but `u64::MAX` makes a tx inserted during clock skew look newer than every other unconfirmed tx. Using `0` seems safer for conflict ordering and matches the idea that we could not get a trustworthy timestamp.

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -439,6 +439,53 @@ impl WalletActor {
...
+                // wallet, the only remaining work is session cleanup — no network needed.
+                // Attempt it here so this send can proceed without requiring a restart.
+                let persister = PayjoinSessionPersister::new(self.db.clone());
+                let can_cleanup = persister
+                    .pending_txid()
+                    .is_some_and(|txid| self.wallet.bdk.get_tx(txid).is_some());
+
+                if !can_cleanup {
+                    return Produces::ok(Err(Error::SignAndBroadcastError(
+                        "a previous payjoin session is pending recovery; please try again later"
```
praveenperera: Longer-term, this probably needs a user-visible recovery/abandon path. If the committed terminal tx is permanently rejected by the network, startup will keep retrying, `get_transaction` will keep saying it is not known, and this gate leaves the wallet unable to send with no way out other than manual database surgery.

rust/src/database/wallet_data.rs
```diff
@@ -79,6 +81,24 @@ impl ReceiveAddressCache {
     }
 }
 
+/// Terminal action decided for a payjoin session before the broadcast completes
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub enum PendingAction {
+    BroadcastFallback,
+    BroadcastProposal { transaction: Vec<u8> },
```
praveenperera: why is the transaction bytes here? don't we usually carry it around as a typed transaction?

praveenperera (changes requested):

rust/src/manager/wallet_manager/actor/transactions.rs
Sandipmandal25: will track this as a follow up for a cancel button.

rust/src/database/wallet_data.rs
Sandipmandal25: We store the transaction as raw bytes because bitcoin::Transaction doesnt implement serde which is what the database uses. so encoding it ourselves keeps the stored bytes identical to what gets broadcast on recovery

Sandipmandal25:
@greptileai

rust/src/manager/wallet_manager/actor/transactions.rs
Sandipmandal25: added!

rust/src/database/wallet_data.rs
praveenperera: @Sandipmandal25  lets create a wrapper new type like `TransactionBytes(Vec<u8>)` we do this a lot in the code, helps not get confused.

Sandipmandal25: done it!

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -454,6 +454,53 @@ impl WalletActor {
...
+            Ok(Some(_)) => {
+                // If the session has a terminal marker and the tx is already in the local
+                // wallet, the only remaining work is session cleanup — no network needed.
+                // Attempt it here so this send can proceed without requiring a restart.
+                let persister = PayjoinSessionPersister::new(self.db.clone());
+                let can_cleanup = persister
+                    .pending_txid()
+                    .is_some_and(|txid| self.wallet.bdk.get_tx(txid).is_some());
+
+                if !can_cleanup {
```
praveenperera, line 469: On restart this should retry through `resume_payjoin_session()`, but I do not see an in-process retry after a transient node failure. `handle_payjoin_terminal_broadcast_result` clears `payjoin_actor`, then the next send hits this gate and returns instead of re-dispatching the stored fallback/proposal.

Can we retry the pending terminal action here, or add an explicit retry path?

rust/src/database/wallet_data.rs
```diff
@@ -79,6 +81,39 @@ impl ReceiveAddressCache {
...
+#[derive(
+    Debug, Clone, Serialize, Deserialize, Eq, PartialEq, derive_more::From, derive_more::Into,
+)]
+pub struct TransactionBytes(Vec<u8>);
+
+impl AsRef<[u8]> for TransactionBytes {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
```
praveenperera: i think `derive_more` as an `AsRef`

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -713,25 +843,70 @@ impl WalletActor {
...
+    ) -> ActorResult<()> {
+        let proposal_tx = match self.do_sign_original_psbt(proposal_psbt).await {
+            Ok((_, tx)) => tx,
+            Err(error) => {
+                error!("failed to sign recovered payjoin proposal, pausing for retry: {error:?}");
+                // the receiver accepted a valid proposal; do not fall back — retain the record
+                // so the user can retry after resolving the signing failure
+                self
+                    .send(WalletManagerReconcileMessage::WalletError(Error::SignAndBroadcastError(
+                    "could not sign the recovered payjoin proposal; unlock the wallet and restart"
```
praveenperera, line 889: i'm seing more and more string error messages, i wonder if this means we need to make `SignAndBroadcastError` better, maybe it should be a nested error enum?

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -905,6 +1080,62 @@ async fn broadcast_transaction_with_connection(
...
+    Ok(())
+}
+
+async fn payjoin_tx_known_to_node(addr: WeakAddr<WalletActor>, txid: Txid) -> bool {
+    let node_client = match call!(addr.node_client_for_broadcast()).await {
+        Ok(Ok(node_client)) => node_client,
+        _ => return false,
+    };
+
+    matches!(node_client.get_transaction(txid).await, Ok(Some(ref found)) if found.compute_txid() == txid)
```
praveenperera: this code is a nice one-liner, but splitting it up would make it easier to read

praveenperera (changes requested):

rust/src/manager/wallet_manager/actor/transactions.rs
Sandipmandal25: agree! should i handle it here? so should i create a issue ?

rust/src/manager/wallet_manager/actor/transactions.rs
Sandipmandal25: retries the pending action inline now instead of waiting for a restart

Sandipmandal25:
@greptileai

rust/src/manager/wallet_manager/actor/transactions.rs
praveenperera: create an issue

Sandipmandal25: Ok sure! Creating an issue for this

Sandipmandal25: #836 is related to this.

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -478,6 +478,55 @@ impl WalletActor {
...
+                }
+
+                if let Err(error) = self.db.delete_payjoin_sender_session() {
+                    warn!("payjoin session cleanup at send gate failed: {error}");
+                    return Produces::ok(Err(Error::SignAndBroadcastError(
+                        "a previous payjoin session is pending cleanup; please try again later"
+                            .to_string(),
+                    )));
+                }
+                // record cleared — fall through to allow this send
```
praveenperera: Could we avoid falling through into the new send after cleaning up the previous session here? If the terminal transaction was broadcast but session deletion failed, both clients show an error and let the user retry the same confirmation. That retry can clear the record here and then immediately start the supplied PSBT again. When the completed transaction was the payjoin proposal, the supplied PSBT is still the original fallback, so it can conflict with or replace the proposal we already broadcast. I think cleanup completion needs an early/typed outcome so this request is not consumed, or cleanup should happen before the send UI is enabled and require the current payment to be confirmed again.

praveenperera (changes requested):
almost there

rust/src/manager/wallet_manager/actor/transactions.rs
Sandipmandal25: cleanup now returns early so the supplied psbt is never consumed. The user sees an error and re confirms from scratch

Sandipmandal25:
@greptileai

PR #826, opened 2026-07-06
Refactor cloud backup and mobile manager ownership

praveenperera:


praveenperera:
@greptile-apps

praveenperera:
@greptile-apps thanks for the review. I looked through these, and I do not think any of the three cloud-backup findings should block this refactor.

1. **Bounded reconcile send**

This is a theoretical failure mode, but not a realistic PR blocker. The cloud-backup reconcile queue has capacity 1000 and is drained by a dedicated listener. Lifecycle messages are emitted on state transitions, not as a high-volume event stream, so organically filling that queue would require the platform-side consumer to be wedged. If the consumer is wedged badly enough to stop draining 1000 messages, the app is already broken; changing this specific send path to async would not make cloud backup healthy. This behavior is also not introduced by this refactor: `master` already used a bounded `flume::Sender::send` for the same cloud-backup reconcile messages.

2. **Partial multi-file Drive delete**

The delete operation is idempotent and retryable. If deleting one location succeeds and a later location fails, the next attempt only finds the remaining files and continues cleanup. I do not see a caller that treats the aggregate error as meaning "nothing was deleted"; callers treat it as a failed delete attempt, which is compatible with best-effort remote cleanup. This behavior also existed before the refactor; the refactor moved the Drive code into helpers but did not introduce the delete semantics.

3. **Namespace validation before Android delete**

Production cloud-backup deletes go through Rust `CloudStorageClient::delete_namespace`, which validates namespace IDs before calling the platform implementation. The direct Kotlin callers are debug wipe utilities, and they source namespaces from `listNamespaces`, which is already filtered to valid backup namespace IDs through the Rust wrapper. Android's resolver also deletes by exact folder-name match and returns not-found if no folder matches, so a malformed string does not create a path to delete an unrelated Drive folder. Kotlin-side validation would be defense-in-depth only, and this was also pre-existing behavior rather than a refactor regression.

Bottom line: these are acknowledged as theoretical hardening notes, but they are pre-existing or low-risk behaviors and should not delay this refactor without a concrete reachable failure scenario.

PR #838, opened 2026-07-15
Gate payjoin at the rust layer

Sandipmandal25:
## Summary

PayJoin is not yet ready for production. This PR temporarily disables PayJoin at the Rust layer by ignoring any provided PayJoin endpoint during the send flow. If a PayJoin endpoint is present, the wallet falls back to a standard transaction broadcast.

The existing `initiate_payjoin_payment` implementation is retained and marked with `#[allow(dead_code)]`, allowing PayJoin to be re enabled in the future by wiring the endpoint back into the send flow without rewriting the underlying logic.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified payment initiation to sign and broadcast transactions directly.
  * Improved payment reliability by removing conditional payjoin routing from the standard payment flow.

* **Maintenance**
  * Retained payjoin components for future use without affecting regular payment processing.

### DIFF
--- a/rust/src/manager/wallet_manager/actor/transactions.rs
+++ b/rust/src/manager/wallet_manager/actor/transactions.rs
@@ -472,10 +472,11 @@ impl WalletActor {
             return Produces::ok(Err(error));
         }
 
-        let (_, transaction) = match self.do_sign_original_psbt(psbt).await {
-            Ok(result) => result,
-            Err(error) => return Produces::ok(Err(error)),
-        };
+        if self.payjoin_actor.is_some() {
+            return Produces::ok(Err(Error::SignAndBroadcastError(
+                "a payjoin session is already in progress".to_string(),
+            )));
+        }
 
         match self.db.get_payjoin_sender_session() {
             Ok(None) => {}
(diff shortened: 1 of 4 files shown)


Sandipmandal25:
@coderabbitai review

rust/src/manager/wallet_manager/actor/transactions.rs
```diff
@@ -466,34 +466,21 @@ impl WalletActor {
     pub async fn initiate_payment(
         &mut self,
         psbt: Psbt,
-        payjoin_endpoint: Option<String>,
+        _payjoin_endpoint: Option<String>,
```
praveenperera: Could we also clear the Payjoin endpoint before building the confirmation route while Payjoin is gated? Right now this path broadcasts the original transaction, but both iOS and Android still see a non-nil endpoint and stay in their sending state waiting for `PayjoinTxBroadcast`, which this path no longer emits. That leaves the transaction successfully broadcast while the UI keeps spinning and authentication stays locked. It would also be helpful to add a regression test covering a Payjoin URI completing through the gated fallback path.

praveenperera (changes requested):

Sandipmandal25:
@greptileai

rust/src/manager/wallet_manager/actor/transactions.rs
Sandipmandal25: cleared the endpoint before building the confirmation route and added a regression test for the fallback path.

Sandipmandal25:
@greptileai

Sandipmandal25:
@praveenperera solved the conflicts and rebased!

Issue #776, opened 2026-05-30
Local certificate store is not used


luisschwab:
OS: `Android`
Cove Version: `1.2.5`

I have a self-hosted CA authority in order to have HTTPS with my internal DNS. My device trusts the certificate chain, other services work just fine.

Cove appears to not be querying the local CA store, as I'm getting this error:

```
Connection Failed

Failed to connect to node
failed to check node url: failed to connect to node: Made one or multiple attempts, all errored:
- invalid peer certificate: Unknown Issuer
```

I've tested this on Nunchuck and it works as expected.

This issue probably stems from either `reqwest` or `bdk_electrum`.

praveenperera:
@luisschwab i think I'd need to fork electrum-client for this

PR #843, opened 2026-07-19
Support self-signed Electrum certificates

### LINKED ISSUE #298: Can't connect to fulcrum server
I can sucessfully connect to my local electrs server using tcp://<ip>:<port> 

However connecting to my Fulcrum server using ssl://<ip>:<port> gives the error 'Something went wrong...'

I want to note I have connected to the same fulcrum server on the same client device using bluewallet so it's not an issue with the server.

Also having a more clear ssl toggle would be nice rather than having to add it to the URL which definitely feels more clumsy than necessary.

kwsantiago:
Closes #298.

A custom Electrum node could only be reached over a chain anchored in the bundled webpki roots, so a self hosted server was unreachable unless its certificate came from a public CA.

`Node` gains an optional `TlsTrust`. Unset keeps today's behavior exactly. Saving a custom node whose certificate cannot be verified now shows that certificate's SHA-256 fingerprint and asks whether to trust it.

Two rules make it opt in without adding a setting:

- The prompt appears only after normal verification has already failed, so it is never put to a user who does not have to answer it, and never for preset nodes.
- A url that already trusts a certificate is never offered a different one, it is told the certificate changed. Otherwise trust on first use becomes trust on every save. That decision lives in `certificate_decision` rather than in each app, so both platforms apply the same rule and it can be tested.

Pinning still verifies the handshake signature, so the peer must hold the pinned key. It authenticates against a certificate rather than a name, which is not the same as skipping verification.

## Notes

- `electrum-client` only takes a caller supplied rustls session through `RawClient`, which has no reconnect loop, so `Transport` keeps what it needs to rebuild the connection. The client is cached for a wallet's lifetime, so without that one dropped socket would break a pinned node until restart.
- Esplora and non-`ssl://` urls refuse certificate settings rather than connecting with weaker trust than was asked for.
- Hosts may be IP addresses; an IPv6 literal was rejected outright before.
- `TlsTrust::CustomCa` validates against a user supplied CA and still checks the hostname, so a privately issued leaf can rotate without re-pinning. It is tested, but only the fingerprint mode has UI here since that is what the issue asks for and it needs no file picker. Happy to wire up the paste-a-certificate field as a follow-up, or drop the variant if you would rather it landed with its own UI.
- No new runtime dependencies; `rcgen` is dev only.
- Not #776, which wants the device's CA store and needs `rustls-platform-verifier` or equivalent.

## Testing

1617 Rust tests, `fmt` and `clippy -D warnings` clean. Tests cover the pinned and CA paths, hostname mismatch by IP, IPv6, leaf rotation, that default trust still rejects self-signed, that a rejected certificate is distinguished from an unreachable node, and reconnect including concurrent callers.

Verified on a Pixel 9a against a self-signed front end to a real Electrum server:

| Step | Server | App |
| --- | --- | --- |
| default trust | `REJECTED: UnknownCA` | rejected |
| reading the certificate | `OK, 0 requests` | no traffic carried |
| fingerprint shown | | matches the server's |
| accepting | `OK, 1 request` | connects |
| a different host | `OK, 0 requests` | offers that host's fingerprint |
| wallet sync | `OK, 62 requests` | scan completes |
| saving again | `OK, 1 request` | no prompt |
| restart, reissued certificate | `REJECTED: AccessDenied` | pin enforced |

The last row used no interaction: the app was relaunched and loaded the pin itself.

The Android flow above was exercised on device. The iOS side compiles but has not been run on a device, so the flow itself is unverified there.

Bindings are regenerated here; say the word if you would rather they came from the Regenerate Bindings workflow.

## Known limitations

- Cove stores one selected node, so selecting a preset and returning loses the pin and asks again. Keeping trust per url is a larger change than this PR.
- No way to forget a trusted certificate, so a server that legitimately reissues one has to be re-added. Obvious follow-up.
- The other asks in #298, an SSL toggle, a connection indicator and an offline mode, are out of scope.


rust/src/node/client/electrum/transport.rs
```diff
@@ -0,0 +1,677 @@
...
+fn connect(url: &str, trust: &TlsTrust) -> Result<RawClient<ElectrumSslStream>, ConnectError> {
+    let (host, port) = target(url)?;
+
+    // A pinned fingerprint ignores this name, but rustls still requires a
+    // syntactically valid one to open the session.
+    let server_name = ServerName::try_from(host.as_str())
+        .map_err(|_| ConnectError::InvalidHost(host.clone()))?
+        .to_owned();
+
+    handshake(server_name, tls::client_config(trust)?, &host, port).map(RawClient::from)
```
praveenperera: Could we negotiate the Electrum protocol version here instead of constructing the client directly with `RawClient::from`? That only wraps the TLS stream and leaves the negotiated version unset. Electrum 1.6 expects `server.version` before other requests, and the client also uses the negotiated version to decode some responses. We need to do this on reconnect as well, otherwise pinned/custom-CA nodes may connect successfully over TLS but fail during the connection check or wallet scan.

ios/Cove/Flows/SettingsFlow/SettingsScreen/NodeSelectionView.swift
```diff
@@ -109,7 +138,12 @@ struct NodeSelectionView: View {
         var node: Node? = nil
 
         do {
-            node = try nodeSelector.parseCustomNode(url: customUrl, name: selectedNodeName, enteredName: customNodeName)
+            node = try nodeSelector.parseCustomNode(
+                url: customUrl,
+                name: selectedNodeName,
+                enteredName: customNodeName,
+                tls: customTls
```
praveenperera: Could we make sure the saved TLS trust is only carried forward when the URL and API type still match the saved pinned Electrum node? Right now, editing the URL keeps the previous certificate pin, so the new endpoint gets checked against the old node certificate and may trigger an unnecessary trust prompt. The same stale value is also passed when switching to Custom Esplora, which prevents the node from being saved because Esplora rejects custom certificate settings. This looks like it affects both the iOS and Android settings screens.

praveenperera:
also bring these over https://github.com/bitcoinppl/cove/commit/3607cf38360e8f84e0e1b93e12418c3f8b3dffa4

rust/src/node/client/electrum/transport.rs
kwsantiago: Done. Negotiating `server.version` on connect and on reconnect, keeping the version per connection.

One thing your comment did not flag: with the version unset, `block_headers` was also decoded as 1.4, which would have failed the scan against a 1.6 server. `Transport::block_headers` now decodes by the negotiated version. Test server answers `server.version` and returns the 1.6 shape.

ios/Cove/Flows/SettingsFlow/SettingsScreen/NodeSelectionView.swift
kwsantiago: Done. Fixed in Rust so both platforms share the rule: `parse_custom_node` inherits the saved trust itself, only for an unchanged url and only for Electrum. The screens now keep only a certificate accepted in the current session, paired with the url it was accepted for.

PR #869, opened 2026-08-06
Add cancellable Payjoin polling countdown

Sandipmandal25:
## Summary

* Add a PayJoin waiting state with a live session countdown on iOS and Android.
* Show when the sender is waiting for the receiver.
* Allow users to cancel PayJoin and send the fallback transaction normally.
* Handle successful broadcasts and errors from the new waiting state.

## Why

PayJoin polling can take time, but the send screen previously showed only a generic loading state with no indication of what was happening or how long remained.

This gives users clearer feedback and lets them fall back to a normal send without waiting for the session to expire.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a payjoin waiting state with a visible countdown during transaction processing.
  * Added an option to cancel an active payjoin attempt.
  * Cancellation now safely falls back and refreshes wallet activity.
* **Bug Fixes**
  * Improved handling of payjoin success and errors while waiting.
  * Countdown state is cleared after broadcasts or send-flow errors.

### DIFF
--- a/android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
+++ b/android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
@@ -18,7 +18,6 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
-import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -45,6 +44,7 @@ sealed interface SendState {
     data object Sending : SendState
 
     data class PayjoinWaiting(
+        val sessionId: PayjoinSessionId,
         val deadlineSecs: ULong,
     ) : SendState
 
@@ -53,6 +53,18 @@ sealed interface SendState {
     data class Error(
         val message: String,
     ) : SendState
+}
+
+private sealed interface SendConfirmationAlert {
+    data object Sent : SendConfirmationAlert
+
+    data class SendFailed(
+        val message: String,
+    ) : SendConfirmationAlert
+
+    data class PayjoinCancellationFailed(
+        val message: String,
+    ) : SendConfirmationAlert
 }
 
 /** send flow container - manages WalletManager + SendFlowManager lifecycle */
@@ -457,13 +469,12 @@ private fun SendFlowRouteToScreen(
         is SendRoute.Confirm -> {
             val details = sendRoute.v1.details
             val input = sendRoute.v1.input
-            val payjoinEndpoint = sendRoute.v1.payjoinEndpoint
+            val unsignedMode = (input as? SendConfirmationInput.Unsigned)?.mode
+            val payjoinIntent = (unsignedMode as? UnsignedPaymentMode.Payjoin)?.intent
 
             var sendState by remember { mutableStateOf<SendState>(SendState.Idle) }
             var finalizedTransaction by remember { mutableStateOf<BitcoinTransaction?>(null) }
-            var showSuccessAlert by remember { mutableStateOf(false) }
-            var showErrorAlert by remember { mutableStateOf(false) }
-            var payjoinInitiated by rememberSaveable { mutableStateOf(false) }
+            var confirmationAlert by remember { mutableStateOf<SendConfirmationAlert?>(null) }
             val scope = rememberCoroutineScope()
 
             // lock on appear for hot wallets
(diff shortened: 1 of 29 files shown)


Sandipmandal25:
@greptileai review

ios/Cove/WalletManager.swift
```diff
@@ -564,12 +568,17 @@ private struct WalletManagerBootstrap {
 
         case let .sendFlowError(error):
             sendFlowErrorAlert = TaggedItem(error)
+            payjoinDeadlineSecs = nil
```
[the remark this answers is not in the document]
Sandipmandal25: close_session is idempotent so only one of success or cancel can fire. the loser is a silent no op. both paths end with the broadcast message that clears the deadline.

Sandipmandal25:
@greptileai re review

Sandipmandal25:
@coderabbitai review

rust/src/manager/wallet_manager/payjoin.rs
```diff
@@ -578,6 +578,13 @@ impl PayjoinActor {
 
         self.poll_deadline = Some(Instant::now() + PAYJOIN_SESSION_TIMEOUT);
         self.session = PayjoinSession::Polling { polling_sender };
+
+        let deadline_secs = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.as_secs() + PAYJOIN_SESSION_TIMEOUT.as_secs())
+            .unwrap_or(0);
+        send!(self.wallet_addr.notify_payjoin_polling_started(deadline_secs));
```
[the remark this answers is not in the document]
Sandipmandal25: already handled. the fallback persistence failure routes through the send flow error channel which clears the deadline on both platforms.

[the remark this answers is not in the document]
Sandipmandal25: `notify_payjoin_error` only fires on session resume where the deadline is already nil from fresh app state. the fallback persistence failure during an active session routes through the send flow error channel not through this handler.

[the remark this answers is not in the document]
Sandipmandal25: i was wrong fixed do rechcek it is now also cleared in the wallet error handler on both platforms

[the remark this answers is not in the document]
Sandipmandal25: fixed do rechcek again

Sandipmandal25:
@praveenperera rebased this pr!

Sandipmandal25:
@greptileai review

ios/Cove/Flows/SendFlow/SendFlowConfirmScreen.swift
```diff
@@ -101,21 +108,39 @@ struct SendFlowConfirmScreen: View {
...
 
         sendState = .sent
         presenter.confirmationAlertState = .init(.sent(id))
         auth.unlock()
     }
 
+    private func cancelPayjoin() {
+        sendState = .sending
+        Task {
+            try? await manager.rust.cancelPayjoin()
```
praveenperera: The iOS build currently fails because `WalletManager.rust` is a private optional property, so this screen cannot call `cancelPayjoin()` directly. Could you add a `WalletManager.cancelPayjoin()` wrapper that uses `withRustAsync`, and call that wrapper from the screen? This also keeps the code aligned with `ARCHITECTURE.md:177`.

rust/src/manager/wallet_manager/actor.rs
```diff
@@ -470,7 +470,21 @@ impl WalletActor {
...
-        self.send(WalletManagerReconcileMessage::WalletError(Error::PayjoinSessionError(msg)));
+        self.send(WalletManagerReconcileMessage::SendFlowError(
+            SendFlowErrorAlert::SignAndBroadcast(msg),
+        ));
+        Produces::ok(())
+    }
+
+    pub async fn cancel_payjoin(&mut self) -> ActorResult<Result<(), Error>> {
+        if let Some(actor) = self.payjoin_actor.take() {
+            send!(actor.cancel_and_fallback());
```
praveenperera: When an active session is cancelled, `PayjoinActor::cancel_and_fallback` returns `Some(fallback_tx)` after persisting the fallback marker and closing the child. The current fire-and-forget call drops that transaction, so the fallback is not broadcast and no `PayjoinTxBroadcast` is emitted. Both screens can remain in `Sending` until the app restarts. Could the child result be awaited and passed to `handle_payjoin_fallback`? `ARCHITECTURE.md:58` also requires the manager to apply child results.

rust/src/manager/wallet_manager.rs
```diff
@@ -823,6 +824,13 @@ impl RustWalletManager {
         Ok(())
     }
 
+    #[uniffi::method]
+    pub async fn cancel_payjoin(&self) -> Result<(), Error> {
+        call!(self.actor.cancel_payjoin()).await.unwrap()?;
+        self.force_wallet_scan().await;
```
praveenperera: `force_wallet_scan()` returns a `Result`, but this statement ignores it. Since Rust CI runs Clippy with `-D warnings` (`.github/workflows/ci.yml:210`), `unused_must_use` will fail the build. Please either propagate the result or explicitly handle the failure if this is intended to be best effort.

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
```diff
@@ -468,10 +472,21 @@ private fun SendFlowRouteToScreen(
                 }
             }
 
+            // restore or transition to waiting UI from manager-owned deadline;
+            // survives Compose recreation on config changes like rotation
+            LaunchedEffect(walletManager.payjoinDeadlineSecs) {
```
praveenperera: If a persisted Payjoin resumes before navigation, or the user leaves its confirmation and opens another one, the app-owned manager still has the old deadline. This effect can then mark any new confirmation for that wallet as `PayjoinWaiting`, even when its endpoint and details belong to a different payment. The banner shows one payment while Cancel acts on the old session. Could the session or transaction identity be carried through, so the deadline is restored only for the matching confirmation? This is required by `AGENTS.md:12`.

ios/Cove/Flows/SendFlow/SendFlowConfirmScreen.swift
```diff
@@ -101,21 +108,39 @@ struct SendFlowConfirmScreen: View {
         if sinceLocked < 5 { auth.lockState = .unlocked }
     }
 
+    private func payjoinPollingStarted(_: UInt64?, _ deadline: UInt64?) {
+        guard let deadline, case .sending = sendState else { return }
```
praveenperera: When `payjoinDeadlineSecs` is already set before this view appears, for example after a persisted polling session resumes, `onChange` does not run on the first render by default. The new view starts in `.idle`, so the guard rejects the existing deadline. The iOS countdown and cancel action then never appear for the resumed session. Could the initial matching deadline be applied separately from the change handler?

ios/Cove/Flows/SendFlow/ConfirmScreen/SwipeToSendView.swift
```diff
@@ -207,7 +208,7 @@ private struct SwipeToSendStatus: View {
             switch sendState {
             case .idle:
                 EmptyView()
-            case .sending:
+            case .sending, .payjoinWaiting:
```
praveenperera, line 213: While the state is `.payjoinWaiting`, `SwipeToSendHandle` still has its drag gesture, and `dragEnded` calls `onConfirm()` without checking the state. A user can swipe again during the ten-minute wait, which starts `initiatePayment` again and replaces the waiting UI with the active-session error while the original Payjoin continues. Could the gesture be enabled only when the state is `.idle`, as it already is on Android?

praveenperera (changes requested):

android/app/src/main/java/org/bitcoinppl/cove/flows/SendFlow/SendFlowContainer.kt
Sandipmandal25: Fixed with a saved local guard so only the same screen that started Payjoin can show the waiting state.

ios/Cove/Flows/SendFlow/ConfirmScreen/SwipeToSendView.swift
Sandipmandal25: users cannot swipe again while Payjoin is waiting

ios/Cove/Flows/SendFlow/SendFlowConfirmScreen.swift
Sandipmandal25: only showing waiting after this screen has already started sending

PR #872, opened 2026-08-18
Harden wallet lifecycle, restore, TAPSIGNER, and cloud backup

praveenperera:
## Summary

This PR hardens wallet state that can cross an actor, database, NFC, cloud, or app-process boundary.

- Reject persisted fee snapshots whose wall-clock time is in the future. In-process snapshots keep monotonic age.
- Route TAPSIGNER retries through Rust `resolveTapSignerCommand` so a continuation that guards an uncertain card mutation is never dropped. Accept ASCII digit CVCs (6 to 32), validate inputs through one set of Rust rules, and keep NFC sessions cancel-safe on both platforms. Pin `bitcoinppl/rust-cktap` to `3d74272692278f31e2e230b0963a12ebe9b8745b`.
- Cancel discovery scans with one cancellation token that covers the scan loop and the in-flight node connection.
- Make backup restore fail-closed. Decrypt and preflight a backup before destructive cleanup, bind one-use approval to the payload, recover interrupted restores at startup, and keep secret digests out of restore markers.
- Keep cloud backup inventory honest. Treat partial wallet failures as needs-attention with typed issues, let users delete inaccessible cloud-only backups after confirmation, prefer local iCloud reads, and own progress in one place.
- Coordinate wallet deletion and wipe through `WalletLifecycleCoordinator`. The wallet actor owns metadata and publishes full snapshots. Address-type switch publishes the replacement store with one atomic rename and heals forward on load. Delete and wipe wait for construction, persistence, and actors to drain, then return typed retry authorization instead of a force path.
- Bump the iOS build number to 115 and regenerate the Swift and Kotlin UniFFI bindings.

## Why

- A wall-clock change could treat a future persisted fee snapshot as a usable fallback.
- A lost NFC response after a TAPSIGNER mutation could make the app repeat a command whose result was not known. CVC input used hex and did not match the numeric domain on current cards.
- Discovery shutdown could wait on a node connection that the scan start-gate did not cancel.
- Restore reservation checks could reject stale artifacts without a safe recovery path. Markers stored unsalted SHA-256 digests of secrets.
- Cloud backup verify could hide partial wallet failures, and one hung coordinated iCloud read could outlive the whole-file deadline.
- Wallet metadata was cloned under a read lock and written back after database work, which could erase a concurrent update. Address-type switch used an eight-phase journal that could brick startup. Wallet deletion could report success while leaving secrets, or wipe the metadata a retry needs.

## User impact

- Backup restore can explain and request approval before it removes conflicting data.
- Interrupted restore and TAPSIGNER operations have explicit recovery paths.
- TAPSIGNER users enter the 6 to 32 digit CVC as printed.
- Delete and full wipe show a blocked or retry path instead of a silent force.
- Users can delete cloud-only backups that the active key cannot decrypt, after confirmation.
- Wallet rows stay visible while cloud backup verify or inventory is running.

## Testing

- `RUSTUP_TOOLCHAIN=stable just lint-rust`
- `RUSTUP_TOOLCHAIN=stable just build-ios`
- `RUSTUP_TOOLCHAIN=stable just build-android`

### Platform Coverage

- [ ] Tested on iOS device
- [ ] Tested on iOS simulator
- [x] Not tested

## Checklist


praveenperera:
@greptileai

Issue #858, opened 2026-07-28
Move #[cfg(test)] helpers out of production modules


praveenperera:
## Summary

Keep test-only helpers out of production modules. Prefer free functions or helpers inside `#[cfg(test)] mod tests`, not `#[cfg(test)]` methods/fields on production types.

This matches the project rule: test-only functions, types, and modules stay under `mod tests` (or a dedicated `mod test_support`), and production paths stay free of test scaffolding.

## Motivation

On `keyteleport`, we already cleaned up cases like:

- `SenderSession::with_private_key_and_password` → free helper in `mod tests`
- `RustKeyTeleportManager::current_generation` → helper inside `mod tests`

The same pattern still exists on `master` and should be refactored the same way.

### Test-only methods on production impls

| File | Item |
| --- | --- |
| `rust/crates/cove-common/src/logging/capture.rs` | `snapshot_text`, `clear` |
| `rust/src/database/cloud_backup.rs` | `delete` |
| `rust/src/manager/cloud_backup_manager.rs` | `clear_pending_enable_session` |
| `rust/src/manager/cloud_backup_manager/actors/cleanup.rs` | `is_idle_for_test` |
| `rust/src/manager/cloud_backup_manager/actors/supervisor.rs` | `take_claim` |
| `rust/src/manager/cloud_backup_manager/actors/supervisor/enable/discard.rs` | `clear_pending_enable_session` |
| `rust/src/manager/reconcile_channel.rs` | `receiver` |
| `rust/src/manager/wallet_manager/actor/transaction_confirmation.rs` | `#[cfg(test)] impl WalletActor` with `*_for_test` methods |

### Test-only fields / constructors

| File | Item |
| --- | --- |
| `rust/src/manager/cloud_backup_manager/wallets/passkey/authorization_retry.rs` | `jitter_seed` field, `for_test` constructor, and test-only branches |

### Out of scope / intentional

These are fine as-is unless we decide to tighten further:

- `#[cfg(test)] mod tests { ... }` blocks
- `#[cfg(test)] pub(crate) use ...` re-exports of real `test_support` modules
- Production APIs that also happen to be useful in tests (e.g. session restore constructors)

## Desired end state

1. No `#[cfg(test)]` methods or fields on production `impl` blocks outside test modules.
2. Test setup uses free helpers or `impl` extensions defined inside `mod tests` / `test_support`.
3. Production crates compile without test-only surface area in the main API.

## Suggested approach

[...]

PR #883, opened 2026-09-18
refactor unit tests:issue-858

RandyMcMillan:


RandyMcMillan:
[1b461fc](https://github.com/bitcoinppl/cove/commit/1b461fce07e105d6cd666f5544073e9165ed9d31) remove my local dev config

rust/src/manager/cloud_backup_manager/wallets/passkey/authorization_retry.rs
```diff
@@ -259,12 +241,30 @@ mod tests {
...
+            .with_max_delay(config.max_delay)
+            .without_max_times()
+            .with_total_delay(Some(config.total_delay));
+
+        if config.jitter {
+            builder = builder.with_jitter_seed(seed);
+        }
+
+        builder.build().map(move |delay| delay.min(config.max_delay))
+    }
```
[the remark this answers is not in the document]
RandyMcMillan: de-dup test logic here: https://github.com/bitcoinppl/cove/commit/460594d6016f14799be4bee6d4a33f8cef7239d5

rust/src/manager/cloud_backup_manager/store.rs
```diff
@@ -564,14 +565,14 @@ mod tests {
...
-        let _ = db.cloud_backup_state.delete();
+        let _ = delete_cloud_backup_state(&db.cloud_backup_state);
     }
 
     #[test]
     fn reset_verification_persists_pending_completion_in_enabled_state_write() {
         let _guard = setup_database_test();
         let db = Database::global();
-        let _ = db.cloud_backup_state.delete();
+        let _ = delete_cloud_backup_state(&db.cloud_backup_state);
```
praveenperera, line 575: how dont see how this is better?

praveenperera (changes requested):
Some good changes but some churn thats not adding much benefit

rust/src/manager/cloud_backup_manager/store.rs
RandyMcMillan: The rule: test-only functions stay in mod tests / mod test_support, not as #[cfg(test)] methods on production types.                                                                                             


1. Production API hygiene — CloudBackupStateTable::delete was only callable in tests, but it lived on the production impl block. Any engineer reading the production code sees a delete method that doesn't      
      actually exist in production builds. Moving it to test_support makes the boundary explicit.                                                                                                                   
                                                                                                                                                                                                                    
2. Prevent accidental production use — with the method on the type, it's easy for someone to write db.cloud_backup_state.delete() in production code. It would compile in tests but fail in release. The free    
      function test_support::delete(&table) makes the test-only nature obvious at every call site.                                                                                                                  
                                                                                                                                                                                                                    
3. Consistency with the rest of the codebase — the issue already cleaned up the same pattern on keyteleport (SenderSession::with_private_key_and_password → free helper,                                         
      RustKeyTeleportManager::current_generation → helper inside mod tests). This applies the same rule to cloud backup.

rust/src/database/cloud_backup.rs
```diff
@@ -95,21 +95,6 @@ impl CloudBackupStateTable {
 
         Ok(CommittedCloudBackupStateMutation { outcome, state })
     }
-
-    #[cfg(test)]
```
RandyMcMillan, line 99: Keep test-only helpers out of production modules. Prefer free functions or helpers inside #[cfg(test)] mod tests, not #[cfg(test)] methods/fields on production types.

This matches the project rule: test-only functions, types, and modules stay under mod tests (or a dedicated mod test_support), and production paths stay free of test scaffolding.

REF: https://github.com/bitcoinppl/cove/issues/858

rust/src/manager/cloud_backup_manager/store.rs
RandyMcMillan: https://github.com/bitcoinppl/cove/pull/883#discussion_r4047950551

praveenperera: I agree that this helper should live under `test_support`. I do not see why it must become a free function. This PR already keeps inherent test-only methods inside `test_support` for `WalletActor` and `CloudBackupCleanupWorker`. Please move this impl into `test_support` and keep the existing `db.cloud_backup_state.delete()` call sites. That meets issue #858 without the call-site churn.

rust/src/manager/cloud_backup_manager/wallets/passkey/authorization_retry.rs
```diff
@@ -85,22 +83,14 @@ impl PlatformAuthorizationRetrier {
...
-        retrier.jitter_seed = Some(jitter_seed);
-        retrier
+        Self { policy, deadline: Instant::now() + policy.config().total_delay }
     }
 
-    fn retry_backoff(&self, total_delay: Duration) -> impl backon::Backoff {
+    fn retry_backoff(
+        &self,
+        total_delay: Duration,
+        jitter_seed: Option<u64>,
```
praveenperera, line 92: This still adds a test-only input to production code. Production always passes `None`, and only the test passes `Some(seed)`. Removing the `#[cfg(test)]` field does not help if the same test seam moves into the production method signature and branch. Please keep the production backoff path free of this test-only input. Use a test-only helper, or model deterministic jitter as a real dependency if the production design needs it.

rust/src/manager/reconcile_channel.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) use impl_reconcile_sink;
 #[derive(Debug)]
 pub struct ReconcileChannel<M: DebugSend> {
     sender: MessageSender<M>,
-    receiver: Arc<Receiver<SingleOrMany<M>>>,
+    pub(crate) receiver: Arc<Receiver<SingleOrMany<M>>>,
```
praveenperera: Please keep `receiver` private. Making it `pub(crate)` in all builds expands the production surface only so tests can read it. This conflicts with the desired end state in issue #858. Put the test-only access in a `#[cfg(test)]` `test_support` helper instead.

RandyMcMillan:
maybe this should be split into seperate PRs - limit scope to individual PRs

